PR #7, opened 2020-03-27
DR-831 stairway scaling features

ddietterich:
- Extended the database schema to support sharing one database by many Stairway instances
- Implemented graceful shutdown of stairway (quietdown and terminate)
- Implemented yield to free the Stairway thread on a long wait
- Implemented the rerun response to allow "paging" execution
- Added tests for all new functions

### DIFF
--- a/src/main/java/bio/terra/stairway/Flight.java
+++ b/src/main/java/bio/terra/stairway/Flight.java
@@ -161,22 +161,29 @@ public class Flight implements Callable<FlightState> {
                 return result;
             }
 
-            // Record the step state
-            flightDao.step(context());
-
             switch (result.getStepStatus()) {
                 case STEP_RESULT_SUCCESS:
-                    // Run the next step (if any)
+                    // Finished a step; run the next one
+                    context().setRerun(false);
+                    flightDao.step(context());
                     context().nextStepIndex();
                     break;
 
                 case STEP_RESULT_RERUN:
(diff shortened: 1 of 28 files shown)


src/main/java/bio/terra/stairway/Flight.java
```diff
@@ -94,6 +95,12 @@ private FlightStatus fly() {
             if (context().isDoing()) {
                 StepResult doResult = runSteps();
                 if (doResult.isSuccess()) {
+                    if (doResult.getStepStatus() == StepStatus.STEP_RESULT_STOP) {
+                        return FlightStatus.READY;
+                    }
+                    if (doResult.getStepStatus() == StepStatus.STEP_RESULT_YIELD) {
+                        return FlightStatus.WAITING;
+                    }
```
RoriCremer: are we going to want to have more granular results in the job results as well?

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -161,7 +161,7 @@ void submit(FlightContext flightContext) throws DatabaseOperationException {
     }
 
     /**
-     * Record the flight state right before a step
+     * Record the flight state right after a step
```
RoriCremer: Is this just a comment correction, or are we changing how the steps are recorded because of the new shut down options and I'm just missing where we are doing that?

PR #9, opened 2020-04-07
DR-831 stairway scaling

ddietterich:
Accidently merged this, then reverted. So take 2 on the review.

### DIFF
--- a/src/main/java/bio/terra/stairway/Flight.java
+++ b/src/main/java/bio/terra/stairway/Flight.java
@@ -161,22 +161,29 @@ public class Flight implements Callable<FlightState> {
                 return result;
             }
 
-            // Record the step state
-            flightDao.step(context());
-
             switch (result.getStepStatus()) {
                 case STEP_RESULT_SUCCESS:
-                    // Run the next step (if any)
+                    // Finished a step; run the next one
+                    context().setRerun(false);
+                    flightDao.step(context());
                     context().nextStepIndex();
                     break;
 
                 case STEP_RESULT_RERUN:
                     // Rerun the same step
+                    context().setRerun(true);
+                    flightDao.step(context());
                     break;
 
                 case STEP_RESULT_YIELD:
+                    // Finished a step; yield execution
+                    context().setRerun(false);
+                    flightDao.step(context());
+                    return result;
+
                 case STEP_RESULT_STOP:
-                    // Stop executing
+                    // Stop executing - leave rerun setting as is; we'll need to pick up where we left off
+                    flightDao.step(context());
                     return result;
 
                 case STEP_RESULT_FAILURE_RETRY:
--- a/src/main/java/bio/terra/stairway/Flight.java
+++ b/src/main/java/bio/terra/stairway/Flight.java
@@ -78,11 +78,17 @@ public class Flight implements Callable<FlightState> {
      * so we may be headed either direction.
      */
     public FlightState call() throws DatabaseOperationException, FlightException {
-        logger.debug("Executing flight class: " + context().getFlightClassName() + " id: " + context().getFlightId());
-        FlightStatus flightStatus = fly();
-        context().setFlightStatus(flightStatus);
-        flightDao.exit(context());
-        return flightDao.getFlightState(context().getFlightId());
+        try {
+            logger.debug("Executing: " + context().toString());
+            FlightStatus flightStatus = fly();
+            context().setFlightStatus(flightStatus);
+            flightDao.exit(context());
+            return flightDao.getFlightState(context().getFlightId());
+        } catch (Exception ex){
+            // This is really bad news
+            logger.error("Flight failed with unexpected exception: " + ex.toString());
+            throw ex;
+        }
     }
 
     /**
@@ -98,7 +104,7 @@ public class Flight implements Callable<FlightState> {
                     if (doResult.getStepStatus() == StepStatus.STEP_RESULT_STOP) {
                         return FlightStatus.READY;
                     }
-                    if (doResult.getStepStatus() == StepStatus.STEP_RESULT_YIELD) {
+                    if (doResult.getStepStatus() == StepStatus.STEP_RESULT_WAIT) {
                         return FlightStatus.WAITING;
                     }
                     return FlightStatus.SUCCESS;
@@ -175,7 +181,7 @@ public class Flight implements Callable<FlightState> {
                     flightDao.step(context());
                     break;
 
-                case STEP_RESULT_YIELD:
+                case STEP_RESULT_WAIT:
                     // Finished a step; yield execution
                     context().setRerun(false);
                     flightDao.step(context());
@@ -189,16 +195,14 @@ public class Flight implements Callable<FlightState> {
                 case STEP_RESULT_FAILURE_RETRY:
                 case STEP_RESULT_FAILURE_FATAL:
                 default:
-                    // unreachable
-                    break;
+                    throw new StairwayExecutionException("Unexected step status: " + result.getStepStatus());
             }
         }
         return result;
     }
 
     private StepResult stepWithRetry() throws InterruptedException, StairwayExecutionException {
-        logger.debug("Executing flight id: " + context().getFlightId() + " step: " + context().getStepIndex() +
-            " direction: " + (context().isDoing() ? "doing" : "undoing"));
+        logger.debug("Executing " + context().prettyStepState());
 
         StepRetry currentStep = getCurrentStep();
         currentStep.retryRule.initialize();
@@ -224,10 +228,7 @@ public class Flight implements Callable<FlightState> {
             } catch (Exception ex) {
                 // The purpose of this catch is to relieve steps of implementing their own repetitive try-catch
                 // simply to turn exceptions into StepResults.
-                logger.info("Caught exception: (" + ex.toString() +
-                    ")\nexecuting flight id: " + context().getFlightId() +
-                    " step: " + context().getStepIndex() +
-                    " direction: " + (context().isDoing() ? "doing" : "undoing"), ex);
+                logger.info("Caught exception: (" + ex.toString() + ") " + context().prettyStepState());
 
                 StepStatus stepStatus = (ex instanceof RetryException)
                     ? StepStatus.STEP_RESULT_FAILURE_RETRY
@@ -237,6 +238,7 @@ public class Flight implements Callable<FlightState> {
 
             switch (result.getStepStatus()) {
                 case STEP_RESULT_SUCCESS:
+                case STEP_RESULT_RERUN:
                     if (context().getStairway().isQuietingDown()) {
                         // If we are quieting down, we force a stop
                         result = new StepResult(StepStatus.STEP_RESULT_STOP, null);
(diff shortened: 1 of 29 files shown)


src/main/java/bio/terra/stairway/Flight.java
```diff
@@ -154,9 +161,37 @@ private StepResult runSteps() throws InterruptedException, StairwayExecutionExce
...
+                case STEP_RESULT_STOP:
+                    // Stop executing - leave rerun setting as is; we'll need to pick up where we left off
+                    flightDao.step(context());
+                    return result;
 
-            context().nextStepIndex();
+                case STEP_RESULT_FAILURE_RETRY:
+                case STEP_RESULT_FAILURE_FATAL:
+                default:
+                    // unreachable
```
RoriCremer, line 192: because there should always be an enum value?

ddietterich: Because the `if (!result.isSuccess())` return above would have already returned.
I should throw if the Step ever returns a step result that is not in the Enum. I'll add that in.

thathert: On that note, if either of the FAILURE cases get reached here should this throw an exception?

fboulnois: I'm a fan of making even `unreachable` sections `throw`

src/main/java/bio/terra/stairway/StepStatus.java
```diff
@@ -1,7 +1,10 @@
 package bio.terra.stairway;
 
 public enum StepStatus {
-    STEP_RESULT_SUCCESS,
-    STEP_RESULT_FAILURE_RETRY,
-    STEP_RESULT_FAILURE_FATAL;
+    STEP_RESULT_SUCCESS,       // step succeeded, ready for next step
+    STEP_RESULT_YIELD,         // step succeeded and yields the thread; for long waits
```
marikomedlock, line 5: This is the return status that requires someone to "manually" resume the flight, right? If so, "YIELD" is misleading to me because in OS-world, that means that a thread moves to the back of the scheduler's priority queue. It's still going to run, just maybe not right away. And if I've understood this correctly, that's not the case here -- someone has to call Stairway.resume to continue running a yielded flight. Maybe "PAUSE" instead?

src/main/resources/stairway/db/changesets/20200310_scale.yaml
```diff
@@ -0,0 +1,36 @@
...
+            columns:
+              - column:
+                  name: stairway_name
+                  type: text
+                  remarks: externally set name of the stairway instance
+                  constraints:
+                    primaryKey: true
+                    nullable: false
+              - column:
+                  name: stairway_id
```
marikomedlock, line 33: This column is also unique, right? I think adding an explicit "unique" constraint would be helpful.

src/test/java/bio/terra/stairway/ScenarioTest.java
```diff
@@ -131,6 +135,209 @@ public void testUndo() throws Exception {
...
+        assertFalse(quietYet, "Not quiet yet");
+
+        // Wake up the thread; it should exit into READY state
+        TestStopController.setControl(1);
+        // Allow time for the flight thread to wake up and exit
+        TimeUnit.SECONDS.sleep(5);
+
+        FlightState state = stairway.getFlightState(flightId);
+        assertThat("State is ready", state.getFlightStatus(), equalTo(FlightStatus.READY));
+        assertNull(state.getStairwayId(), "Flight is unowned");
```
marikomedlock, line 163: I'm confused about the expected order of events here. At this point, the sleeping flight thread either woke up in time to finish the step inside of the thread pool's termination grace period, or it didn't wake up in time and got an interrupt. In both cases, I think you'd get a STEP_RESULT_STOP, which would turn into a FlightStatus.READY, so that check makes sense. But the current step would be different because in one case it was completed successfully and in one case it was interrupted. Am I following this correctly?

Regardless, I think it would be good to test both cases -- interrupt because step did/not finish within quiet down grace period.

src/main/java/bio/terra/stairway/Flight.java
ddietterich: I changed it to throw.

src/main/java/bio/terra/stairway/Flight.java
```diff
@@ -196,10 +237,24 @@ private StepResult stepWithRetry() throws InterruptedException, StairwayExecutio
...
                 case STEP_RESULT_FAILURE_FATAL:
+                case STEP_RESULT_STOP:
+                case STEP_RESULT_YIELD:
                     return result;
 
                 case STEP_RESULT_FAILURE_RETRY:
+                    if (context().getStairway().isQuietingDown()) {
+                        logger.info("Quieting down: not retrying flight id: " + context().getFlightId() +
+                                " step: " + context().getStepIndex() +
+                                " direction: " + (context().isDoing() ? "doing" : "undoing"));
```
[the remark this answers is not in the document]
ddietterich: Thanks for pointing that out. I made a little method in FlightContext. Everything is more beautiful

src/main/java/bio/terra/stairway/Flight.java
```diff
@@ -196,10 +237,24 @@ private StepResult stepWithRetry() throws InterruptedException, StairwayExecutio
 
             switch (result.getStepStatus()) {
```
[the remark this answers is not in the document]
ddietterich: Good catch, thanks!

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -199,15 +333,64 @@ void delete(String flightId) throws DatabaseOperationException {
...
      */
-    List<FlightContext> recover() throws DatabaseOperationException {
+    List<FlightContext> recover(String stairwayId) throws DatabaseOperationException {
         final String sqlActiveFlights = "SELECT flightid, class_name " +
             " FROM " + FLIGHT_TABLE +
-            " WHERE status = 'RUNNING'";
+            " WHERE status = 'RUNNING' AND stairway_id = :stairwayid";
 
-        final String sqlLastFlightLog = "SELECT working_parameters, step_index, doing," +
+        final String sqlLastFlightLog = "SELECT working_parameters, step_index, doing, rerun," +
```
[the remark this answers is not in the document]
ddietterich: Nope - removed

src/main/resources/stairway/db/changesets/20200310_scale.yaml
ddietterich: OK

src/test/java/bio/terra/stairway/ScenarioTest.java
ddietterich: In this test case, there is never an interrupt. The only way for the flight to get to the ready state is if it correctly detects the QuietDown state and stops itself.

There is a test for the terminate case: `testTerminate()` later in the module.

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -60,23 +64,92 @@ void startClean() throws DatabaseSetupException {
...
+     * @return short UUID for this stairway instance
+     * @throws DatabaseOperationException on database errors
+     */
+    String findOrCreateStairwayInstance(String stairwayName) throws DatabaseOperationException {
+        try (Connection connection = dataSource.getConnection()) {
+            startTransaction(connection);
+
+            String stairwayId = lookupStairwayInstanceQuery(connection, stairwayName);
+            if (stairwayId != null) {
+                return stairwayId;
```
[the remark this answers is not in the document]
ddietterich: I wrote a comment at the top of the DAO describing the convention. I added an explicit startReadOnlyTransaction method and an explicit commitTransaction method. I made a pass adding the right calls in the right places.

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -475,4 +684,10 @@ private void storeInputParameters(Connection connection, String flightId, Flight
         return inputList;
     }
 
+    private void startTransaction(Connection connection) throws SQLException {
```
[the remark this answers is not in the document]
ddietterich: I couldn't easily extend Connection class, because it is an interface. I just used explicit method calls.

src/main/java/bio/terra/stairway/StepStatus.java
ddietterich: Renamed to WAIT.

src/main/java/bio/terra/stairway/StepStatus.java
```diff
@@ -1,7 +1,10 @@
 package bio.terra.stairway;
 
 public enum StepStatus {
-    STEP_RESULT_SUCCESS,
-    STEP_RESULT_FAILURE_RETRY,
-    STEP_RESULT_FAILURE_FATAL;
+    STEP_RESULT_SUCCESS,       // step succeeded, ready for next step
+    STEP_RESULT_WAIT,          // step succeeded; the flight waits for a completion event
```
marikomedlock, line 5: RE the comment: the flight waits for a resume event?

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -29,6 +29,14 @@
  * May want to split this into an interface and an implementation. This implementation
  * has only been tested on Postgres. It may work on other databases, but who knows?
  * <p>
+ * The practice with transaction code is:
+ * <ul>
+ *     <li>Always explicitly start the transaction. That ensures that all transactions are using the
+ *     serializable isolation level.</li>
+ *     <li>Always explicitly declare read only vs write transactions</li>
+ *     <li>Always explicitly commit the transaction. That is not needed when using the resource try,
+ *     but it is useful documentation to see the point where we expect the transaction to be complete.</li>
```
thathert, line 38: 👍 I like this pattern.

src/main/java/bio/terra/stairway/Flight.java
```diff
@@ -78,11 +78,17 @@ protected void addStep(Step step, RetryRule retryRule) {
...
-        context().setFlightStatus(flightStatus);
-        flightDao.exit(context());
-        return flightDao.getFlightState(context().getFlightId());
+        try {
+            logger.debug("Executing: " + context().toString());
+            FlightStatus flightStatus = fly();
+            context().setFlightStatus(flightStatus);
+            flightDao.exit(context());
+            return flightDao.getFlightState(context().getFlightId());
+        } catch (Exception ex){
```
thathert, line 87: nit: missing space between `)` and `{`

src/main/resources/stairway/db/changesets/20200310_scale.yaml
```diff
@@ -0,0 +1,39 @@
...
+                  type: text
+                  constraints:
+                    nullable: false
```
marikomedlock, line 36: Isn't this nullable=true, for when a flight is disowned?

src/test/java/bio/terra/stairway/ScenarioTest.java
marikomedlock: By grace period, I meant the timeout in quietDown->awaitTermination -- I think that throws an interrupt after the timeout expires. How do you know that the sleep step exits before that quietdown timeout?

Yes, I do see the terminate test, looks great.

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -27,11 +29,20 @@
...
```
marikomedlock, line 38: Thanks for writing this out. Looks great.

PR #11, opened 2020-04-14
DR-897 stairway shutdown

ddietterich:
Debug clean shutdown: 
- proper handling of InterruptedException in lots of places.
- Retry loop for serialization errors during the interrupt phase.
- Quick reject of new flights during quiet down
- Bump version to 0.0.5

### DIFF
--- a/src/main/java/bio/terra/stairway/FlightDao.java
+++ b/src/main/java/bio/terra/stairway/FlightDao.java
@@ -5,6 +5,8 @@ import bio.terra.stairway.exception.DatabaseSetupException;
 import bio.terra.stairway.exception.FlightException;
 import bio.terra.stairway.exception.FlightFilterException;
 import bio.terra.stairway.exception.FlightNotFoundException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.sql.DataSource;
 import java.sql.Connection;
@@ -15,6 +17,8 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
 
 /**
  * The general layout of the stairway database tables is:
@@ -39,6 +43,8 @@ import java.util.List;
  * </ul>
  */
 class FlightDao {
+    private static final Logger logger = LoggerFactory.getLogger(FlightDao.class);
+
     static String FLIGHT_TABLE = "flight";
     static String FLIGHT_LOG_TABLE = "flightlog";
     static String FLIGHT_INPUT_TABLE = "flightinput";
@@ -80,7 +86,9 @@ class FlightDao {
      * @return short UUID for this stairway instance
      * @throws DatabaseOperationException on database errors
      */
-    String findOrCreateStairwayInstance(String stairwayName) throws DatabaseOperationException {
+    String findOrCreateStairwayInstance(String stairwayName)
+            throws DatabaseSetupException, DatabaseOperationException {
+
         try (Connection connection = dataSource.getConnection()) {
             startTransaction(connection);
 
@@ -104,12 +112,12 @@ class FlightDao {
                 return stairwayId;
             }
         } catch (SQLException ex) {
-            throw new DatabaseOperationException("Stairway instance find/create failed", ex);
+            throw new DatabaseSetupException("Stairway instance find/create failed", ex);
         }
-
     }
 
-    private String lookupStairwayInstanceQuery(Connection connection, String stairwayName) throws DatabaseOperationException {
+    private String lookupStairwayInstanceQuery(Connection connection, String stairwayName)
+            throws SQLException, DatabaseOperationException {
         final String sqlStairwayInstance = "SELECT stairway_id" +
                 " FROM " + STAIRWAY_INSTANCE_TABLE +
                 " WHERE stairway_name = :stairwayName";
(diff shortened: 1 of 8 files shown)


zloery:
I don't have as much context on stairway internals, but lgtm

src/main/java/bio/terra/stairway/Stairway.java
```diff
@@ -119,17 +125,33 @@ public boolean quietDown(long waitTimeout, TimeUnit unit) {
...
+        List<Runnable> neverStartedFlights = threadPool.shutdownNow();
+        for (Runnable flightRunnable : neverStartedFlights) {
+            Flight flight = (Flight) flightRunnable;
+            try {
+                logger.info("Setting never-started flight ready: " + flight.context().getFlightId());
+                flight.context().setFlightStatus(FlightStatus.READY);
+                flightDao.exit(flight.context());
+            } catch (DatabaseOperationException | FlightException ex) {
+                // Not much to do on termination
+                logger.warn("Unable to exit never-started flight: " + flight);
```
marikomedlock, line 150: Could you include the flightid here, as well? For debugging purposes, just thinking about when I'm querying PostGres.

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -233,28 +239,45 @@ void exit(FlightContext flightContext) throws DatabaseOperationException, Flight
         }
     }
 
+    private static final int RETRIES = 20;
+    private static final int WAIT_MIN = 250;
+    private static final int WAIT_MAX = 1000;
+
+    private void retryWait(String logTag, String flightId) throws InterruptedException {
+        int sleepMS = ThreadLocalRandom.current().nextInt(250, 751);
```
marikomedlock, line 247: I'd not heard of this class before, very useful!

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -281,27 +304,32 @@ private void complete(FlightContext flightContext) throws DatabaseOperationExcep
         String serializedException =
             exceptionSerializer.serialize(flightContext.getResult().getException().orElse(null));
 
-        try (Connection connection = dataSource.getConnection();
-             NamedParameterPreparedStatement statement =
-                 new NamedParameterPreparedStatement(connection, sqlUpdateFlight);
-             NamedParameterPreparedStatement deleteStatement =
-                 new NamedParameterPreparedStatement(connection, sqlDeleteFlightLog)) {
+        for (int retry = 0; retry < RETRIES; retry++) {
```
marikomedlock, line 307: Is the RETRIES here to address possible PostGres serialization exceptions?

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -337,15 +365,15 @@ void delete(String flightId) throws DatabaseOperationException {
             commitTransaction(connection);
 
         } catch (SQLException ex) {
-            throw new DatabaseOperationException("Failed to delete flight", ex);
+            handleSqlException("Failed to delete flight", ex);
```
marikomedlock, line 368: Could you include the flightId in this error message?

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -386,14 +414,15 @@ FlightContext resume(String stairwayId, String flightId) throws DatabaseOperatio
             return flightContext;
 
         } catch (SQLException ex) {
-            throw new DatabaseOperationException("Failed to get flight", ex);
+            handleSqlException("Failed to get flight", ex);
```
marikomedlock, line 417: Same comment as above, could you include the stairway and flight ids in the exception message?

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -703,4 +737,32 @@ private void commitTransaction(Connection connection) throws SQLException {
         connection.commit();
     }
 
+    private void handleSqlException(String message, SQLException ex)
```
marikomedlock, line 740: When you call this method, I think there is often stariway/flight ids passed to the method. I think it would be helpful when debugging if you included those id(s) in the error message. I made this comment a couple places above, but I think the same about all places this method is called.

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -703,4 +737,32 @@ private void commitTransaction(Connection connection) throws SQLException {
...
+        if (ss.equals("40001") || ss.equals("40P01")) {
+            // Serialization or deadlock
+            logger.info("Caught SQL serialization error - retrying");
+            return true;
+        }
+        if (ss.startsWith("08") || ss.startsWith("53")) {
+            // Connection or resource issue
+            logger.info("Caught SQL connection or resource error - retrying");
+            return true;
+        }
```
marikomedlock, line 763: I don't know if there's a Java/JDBC/Spring SQLState enum (brief googling didn't find me anything). Using an "official" enum would be ideal, but barring that could you move these strings into static variables or an enum (of our own)?

marikomedlock:
One more question: Does the base Step class need to add InterruptedException to the method signatures of the do/undo? Just thinking about how we would call a timer for example in a step.

src/main/java/bio/terra/stairway/FlightDao.java
ddietterich: I did not find one either. I will make some constants for these magic numbers.

src/main/java/bio/terra/stairway/FlightDao.java
ddietterich: I changed all uses to provide stairway and flight ids if available, so we will have consistent output.

src/main/java/bio/terra/stairway/FlightDao.java
ddietterich: It is a bummer. Some of the PSQL status codes are there, but not the ones we care about :(
I defined some constants.

ddietterich:
I added InterruptedException to the Step base class methods. It appears that derivations don't have to have that throw clause if they don't actually throw Interrupted.

We will have to make a pass through the code and remove/rethrow all of those places where we currently catch and hide the InterruptedException.

PR #12, opened 2020-04-24
DR-832 stairway queue

ddietterich:
Add a pubsub work queue to Stairway.

**Added Work Queue**
Stairway now creates (or finds) a pubsub queue when it initializes. The pubsub queue is
used to distribute work across multiple Stairway instances. There are three cases where the
queue is used:
1. Explicitly by a caller to cause a flight to be queued instead of run on `this`
instance. We will use this for the bulk load flight to distribute the single file loads
across the DRmanager instances. It uses a new Stairway method: `submitToQueue`.
2. Implicitly by Stairway when `submit` is called and the local threadPool queue is longer
than some K flights.
3. By Stairway when it is shutting down. In progress flights and flights in the threadPool
queue are put on the pubsub work queue.

There are three parts to the implementation.
1. Queue class provides the interface for publishing and receiving messages from a
queue. It doesn't know about message formats or the purpose of the queue.
2. Message format. I implemented a structure for extensible message formats to allow us to
use the work queue for other tasks besides flight scheduling.
  - QueueMessageEnum is an enum of the different message types. There is only one right now.
  - QueueMessageType is a common class that all messages must include. It contains the format version number and the enum of the type of message.
  -  QueueMessageNoFields is a message class that only includes the QueueMessageType. We can deserialize to that to check the version and know what the target class is.
  - QueueMessage is the base class for messages. It knows how to take a message, deserialize it into QueueMessageNoFields, validate the version and enum, and then deserialize it into the correct type.
  - QueueMessageReady is the message that we put in the work queue to say that a flight is
 ready. It contains the flightId to process.
3. WorkQueueListener is a thread that "listens" for work from the work queue. It only
listens for work if the instance threadPool queue is shorter than some K flights. That is,
we don't want to take work from the work queue that we cannot process.

**Breaking Change to Stairway Construction**
There were already three flavors of constructors in Stairway and new parameters keep
coming. I changed the construction to use the Builder pattern. It will be easier to extend
in the future.

**Fixed an undo direction bug**
There was a bug where the undo method of a failing step would not get called if the system
crashed after the failing step wrote its log message, but before the undo happened. To fix
this I changed the `doing` boolean to a `direction` enumeration. It records the SWITCH
state when logging the failing step that is switching from DO to UNDO. Detecting that
state lets us set the correct step index and undo the failing step. That required a change
to the database schema.

**Minor Changes**
- added logback to the test so we get log files from the tests
- added "connected" test category, since the pubsub use requires access to GCP
- tidied up TestUtils


src/main/java/bio/terra/stairway/Stairway.java
```diff
@@ -186,15 +400,29 @@ public void submit(String flightId,
...
+
+        if (workQueueEnabled && shouldQueue) {
+            // Submit to the queue
+            context.setFlightStatus(FlightStatus.READY);
+            flightDao.submit(context);
+
+            String message = QueueMessage.serialize(new QueueMessageReady(context.getFlightId()));
+            workQueue.queueMessage(message);
+            context.setFlightStatus(FlightStatus.QUEUED);
+            flightDao.queued(context);
```
zloery, line 416: These 4 lines are exactly the same as `queueFlight(context);`

src/main/java/bio/terra/stairway/Stairway.java
```diff
@@ -335,6 +562,38 @@ FlightDao getFlightDao() {
...
+    private void queueFlight(FlightContext context)
+            throws DatabaseOperationException, StairwayExecutionException, InterruptedException {
+        // If the flight state is READY, then we put the flight on the queue and mark it queued in the
+        // database. We cannot go directly from RUNNING to QUEUED. Suppose we put the flight in
+        // the queue before it was marked QUEUED. Another Stairway instance would see it was
+        // RUNNING and decide it shouldn't run it. Suppose we put the flight in the queue after
+        // we marked it QUEUED. Then if we fail before we put it on the queue, no one knows it should
+        // be queued. To close that window, we use the READY state to decouple the flight from RUNNING.
+        // Then we queue, then we mark as queued. Another Stairway looking for orphans, will find the READY
+        // and queue it. Putting a flight on the queue twice is not a problem. Stairway instances race to
```
zloery, line 589: To make sure I'm understanding this right: putting a flight on the queue twice (or more) isn't a problem because one of the first steps to accepting a flight from the work queue (before doing any work) is marking ownership in the shared DB, and at-most one instance can do that because DBs are good at transaction management. Is that right?

ddietterich: Exactly right.

ddietterich:
Some answers to Zach's questions:

_Q: Do we need a distinct QUEUED flight state instead of just using READY? It looks like they're treated equivalently for instances reading from the work queue. As far as I can tell, the important distinction comes when exiting a flight: READY flights get explicitly disowned without checking which instance owns them, while QUEUED flights are only best-effort marked as such, since another instance might have picked the flight up in the meantime._

A: When a flight detects that it should shutdown, it sets the flight to the READY state, places the flight into the Work Queue and then sets the flight to the QUEUED state. This is an example of careful writing. Here are the cases to consider:

1. The system fails when the flight is READY, but has not been queued. Recovery is done by another Stairway instance finding READY flights, queuing them, and setting their state to QUEUED.
2. The system fails when the flight is READY and the flight is in the work queue. Another Stairway instance will find the READY flight, queue it, and set its state to QUEUED. At this point, we have duplicate entries in the work queue. That is fine. The Work Queue guarantees at least once delivery, so we use the database state to detect that a Stairway instance has pulled the flight from the queue and is processing it. A second attempt to resume the flight will fail based on the database state.
3. The system fails before the flight is set to READY. We call this the "orphan" case, because the flight still appears to be running on a Stairway instance, but the instance is shut down. Recovery from this case requires monitoring external to Stairway to notice when instances shut down. In Kubernetes, I plan to monitor the pod events from Kubernetes and call a Stairway endpoint when a pod stops or is killed.

The QUEUED state reduces the cases where a Stairway instance will double queue a flight. It is not strictly required, but I think it is useful.

_Q: Do we need explicit disowning, or could both disown() and queued() check that they're the flight owners before changing status? And if we can do that, do we need QUEUED at all?_

A: `disown` and `queued` are two separate state transitions in the careful writing sequence described above. The purpose of `disown` is to remove ownership by a Stairway instance. That leaves the flight in a state where it is safe for another Stairway instance to pick it up. The purpose of `queued` is to document that the flight is in the Work Queue so will be scheduled on another Stairway instance at some point.

Consider the scenario where a Stairway instance pulls a flight from the work queue after it has been published there, but before the `queued` operation is run. In that case, we want the `queued` operation to do nothing.

Suppose we used your suggestion and published the flight to the work queue before we disowned+queued the record in the database. Another Stairway instance could receive the work before we wrote the record. That instance would believe the flight is owned and simply ack the work queue message.

Therefore, we need the flight to be runnable by another Stairway instance (`disown`) before we put it in the work queue. And then we want to mark the flight QUEUED to avoid needless double publishing of the flight into the work queue.

I hope that all makes sense!

src/main/java/bio/terra/stairway/Queue.java
```diff
@@ -0,0 +1,237 @@
...
+    public void queueMessage(String message) throws InterruptedException, StairwayExecutionException {
+        ByteString data = ByteString.copyFromUtf8(message);
+        PubsubMessage pubsubMessage = PubsubMessage.newBuilder().setData(data).build();
+
+        try {
+            // Once published, returns a server-assigned message id (unique within the topic)
+            ApiFuture<String> future = publisher.publish(pubsubMessage);
+            String messageId = future.get();
+            logger.info("Queued message. Id: " + messageId + "; Msg: " + message);
+            } catch (ExecutionException ex) {
```
thathert, line 182: nit: de-indent

src/main/java/bio/terra/stairway/Queue.java
```diff
@@ -0,0 +1,237 @@
...
+    public void purgeQueue() {
+        PullRequest pullRequest =
+                PullRequest.newBuilder()
+                        .setMaxMessages(1)
+                        .setReturnImmediately(true)
+                        .setSubscription(subscriptionName)
+                        .build();
+
+        // Sometimes we get an empty response even when there are messages, so receive empty twice
+        // before calling it purged.
```
thathert, line 196: Is this expected? and is this the recommended way to address the issue?

src/main/java/bio/terra/stairway/Queue.java
```diff
@@ -0,0 +1,237 @@
...
+        }
+    }
+
+    public void deleteQueue() {
+        TopicName topicName = TopicName.ofProjectTopicName(projectId, topicId);
+
+        try (TopicAdminClient topicAdminClient = TopicAdminClient.create()) {
+            topicAdminClient.deleteTopic(topicName);
+        } catch (IOException ex) {
+            logger.warn("Failed to delete topic: " + topicName);
```
thathert, line 227: Include `ex` as second param to help debugging

src/main/java/bio/terra/stairway/Queue.java
```diff
@@ -0,0 +1,237 @@
...
+        }
+
+        try (SubscriptionAdminClient subscriptionAdminClient = SubscriptionAdminClient.create()) {
+            ProjectSubscriptionName subscription = ProjectSubscriptionName.of(projectId, subscriptionId);
+            subscriptionAdminClient.deleteSubscription(subscription);
+        } catch (IOException ex) {
+            logger.warn("Failed to delete subscription: " + subscriptionName);
```
thathert, line 234: Same here

src/main/java/bio/terra/stairway/Stairway.java
```diff
@@ -25,15 +29,127 @@
  * Stairway is the object that drives execution of Flights.
  */
 public class Stairway {
-    private final Logger logger = LoggerFactory.getLogger("bio.terra.stairway");
+    private static final Logger logger = LoggerFactory.getLogger("bio.terra.stairway");
```
thathert, line 32: Switch to class-based instead of String for this

src/test/java/bio/terra/stairway/QueueTest.java
```diff
@@ -0,0 +1,76 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@Tag("connected")
+public class QueueTest {
+    private Logger logger = LoggerFactory.getLogger("bio.terra.stairway");
```
thathert, line 17: nit: move away from string-based

src/test/java/bio/terra/stairway/SleepQueueTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+@Tag("connected")
+public class SleepQueueTest {
+    private Logger logger = LoggerFactory.getLogger("bio.terra.stairway");
```
thathert, line 17: same here

build.gradle
```diff
@@ -15,8 +15,7 @@ plugins {
 }
 
 group 'bio.terra'
-version '0.0.5-SNAPSHOT'
-
+version '0.0.6-SNAPSHOT'
```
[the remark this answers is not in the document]
ddietterich: No, there is nothing to prevent that. Artifactory holds both versions and serves the most recent. We really need some version management scheme for this and the datarepo-client.
I looked at a promising plug-in, but I couldn't figure out how it would fit with our workflow.

I'll make a ticket to track the problem and keep doing background work to find a workable scheme. Its not like we are the only ones who do this!

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -372,52 +401,58 @@ void delete(String flightId) throws DatabaseOperationException, InterruptedExcep
...
                 " SET status = 'RUNNING'," +
                 " stairway_id = :stairwayId" +
                 " WHERE flightid = :flightId";
 
-        try (Connection connection = dataSource.getConnection();
-             NamedParameterPreparedStatement unownedFlightStatement =
-                     new NamedParameterPreparedStatement(connection, sqlUnownedFlight);
-             NamedParameterPreparedStatement takeOwnershipStatement =
-                     new NamedParameterPreparedStatement(connection, sqlTakeOwnership)) {
+        for (int retry = 0; retry < SERIALIZATION_RETRIES; retry++) {
```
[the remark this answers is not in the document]
ddietterich: Yes, true. I incremented myself here. Let me see if a single function interface will work. It would be cleaner.

src/main/java/bio/terra/stairway/Queue.java
ddietterich: The documentation is _very_ unclear on the semantics of the pull. So, no, this is not the recommended way to do it.

The purge is only used in tests to clean out the queue before running another test. I should document that. If we need "purge" in production, deleting the topic and re-creating it would be better.

ddietterich: Now that I write that, what I _should_ do is make the subscription name and subscriber protected and move this to a subclass within the test directory.

src/main/java/bio/terra/stairway/Direction.java
```diff
@@ -0,0 +1,9 @@
+package bio.terra.stairway;
+
+public enum Direction {
```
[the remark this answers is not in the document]
ddietterich: I'm not in the habit of putting type information into names. Is that a norm?

src/test/java/bio/terra/stairway/QueueTest.java
ddietterich: I searched for all of them and fixed them all.

PR #14, opened 2020-05-27
DR-899 add stairway hooks

snf2ye:
The stairway hook mechanism will allow for functions defined in the jade-data-repo to be called during a stairway flight execution. The primary purpose for this functionality will be to allow for:
1. Logging, to measure flight and step execution time
2. Inserting Faults into a flight, allowing greater control and testability of a Stairway flight

This will be merged in conjunction with [this PR](https://github.com/DataBiosphere/jade-data-repo/pull/478).

### DIFF
--- a/src/main/java/bio/terra/stairway/HookWrapper.java
+++ b/src/main/java/bio/terra/stairway/HookWrapper.java
@@ -11,52 +11,13 @@ public class HookWrapper {
     this.stairwayHook = stairwayHook;
   }
 
-  public HookAction startFlight(FlightContext context) {
+  private HookAction handleHook(FlightContext context, HookInterface hookMethod) {
     if (stairwayHook == null) {
       return HookAction.CONTINUE;
     }
     FlightContext contextCopy = makeCopy(context);
     try {
-      return stairwayHook.startFlight(contextCopy);
-    } catch (Exception ex) {
-      logger.info("Stairway Hook failed with exception: {}", ex);
-    }
-    return HookAction.CONTINUE;
-  }
-
-  public HookAction endFlight(FlightContext context) {
-    if (stairwayHook == null) {
-      return HookAction.CONTINUE;
-    }
-    FlightContext contextCopy = makeCopy(context);
-    try {
-      return stairwayHook.endFlight(contextCopy);
-    } catch (Exception ex) {
-      logger.info("Stairway Hook failed with exception: {}", ex);
-    }
-    return HookAction.CONTINUE;
-  }
-
-  public HookAction startStep(FlightContext context) {
-    if (stairwayHook == null) {
-      return HookAction.CONTINUE;
-    }
-    FlightContext contextCopy = makeCopy(context);
-    try {
-      return stairwayHook.startStep(contextCopy);
-    } catch (Exception ex) {
-      logger.info("Stairway Hook failed with exception: {}", ex);
-    }
-    return HookAction.CONTINUE;
-  }
-
-  public HookAction endStep(FlightContext context) {
-    if (stairwayHook == null) {
-      return HookAction.CONTINUE;
-    }
-    FlightContext contextCopy = makeCopy(context);
-    try {
-      return stairwayHook.endStep(contextCopy);
+      return hookMethod.hook(contextCopy);
     } catch (Exception ex) {
       logger.info("Stairway Hook failed with exception: {}", ex);
     }
@@ -73,5 +34,51 @@ public class HookWrapper {
     fc_new.setResult(fc.getResult());
     fc_new.setRerun(fc.isRerun());
     return fc_new;
+  }
+
+  // Hooks
+  public HookAction startFlight(FlightContext context) {
+    return handleHook(context, new StartFlightCommand());
+  }
+
+  public HookAction startStep(FlightContext context) {
+    return handleHook(context, new StartStepCommand());
+  }
+
+  public HookAction endStep(FlightContext context) {
+    return handleHook(context, new EndStepCommand());
+  }
+
+  public HookAction endFlight(FlightContext context) {
+    return handleHook(context, new EndFlightCommand());
+  }
+
+  // Specify hook action
+  private interface HookInterface {
+    HookAction hook(FlightContext context) throws InterruptedException;
+  }
+
+  private class StartFlightCommand implements HookInterface {
+    public HookAction hook(FlightContext context) throws InterruptedException {
+      return stairwayHook.startFlight(context);
+    }
+  }
+
+  private class StartStepCommand implements HookInterface {
+    public HookAction hook(FlightContext context) throws InterruptedException {
+      return stairwayHook.startStep(context);
+    }
+  }
+
+  private class EndStepCommand implements HookInterface {
+    public HookAction hook(FlightContext context) throws InterruptedException {
+      return stairwayHook.endStep(context);
+    }
+  }
+
+  private class EndFlightCommand implements HookInterface {
+    public HookAction hook(FlightContext context) throws InterruptedException {
+      return stairwayHook.endFlight(context);
+    }
   }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/bio/terra/stairway/HookWrapper.java
```diff
@@ -0,0 +1,70 @@
...
+    return handleHook(context, HookType.startStep);
+  }
+
+  public HookAction endStep(FlightContext context) {
+    return handleHook(context, HookType.endStep);
+  }
+
+  private HookAction handleHook(FlightContext context, HookType type) {
+    if (stairway.getStairwayHook() == null) {
+      return HookAction.CONTINUE;
```
snf2ye, line 32: I wasn't sure if we would would this to return fault or continue. I was thinking that we only wanted "fault" to come from the return of the hook, not to signal an error within stairway.

src/main/java/bio/terra/stairway/Stairway.java
```diff
@@ -189,6 +201,37 @@ public Stairway(Stairway.Builder builder) {
     this.projectId = builder.projectId;
     this.workQueueEnabled = (projectId != null);
     this.quietingDown = new AtomicBoolean();
+    this.stairwayHook =
+        (builder.stairwayHook == null) ? populateDefaultStairwayHook() : builder.stairwayHook;
+  }
+
+  public StairwayHook populateDefaultStairwayHook() {
```
snf2ye, line 208: Again, organization - not sure if this is the best place for this chunk of code.

src/main/java/bio/terra/stairway/HookWrapper.java
```diff
@@ -0,0 +1,70 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class HookWrapper {
+  private static final Logger logger = LoggerFactory.getLogger(HookWrapper.class);
+  private Stairway stairway;
+
+  public HookWrapper(Stairway stairway) {
+    this.stairway = stairway;
```
fboulnois, line 11: Should we create a copy or make stairway read-only? I don't think we want to modify the original stairway object accidentally.

src/main/java/bio/terra/stairway/HookWrapper.java
```diff
@@ -0,0 +1,70 @@
...
+
+  public HookAction startStep(FlightContext context) {
+    return handleHook(context, HookType.startStep);
+  }
+
+  public HookAction endStep(FlightContext context) {
+    return handleHook(context, HookType.endStep);
+  }
+
+  private HookAction handleHook(FlightContext context, HookType type) {
```
fboulnois, line 30: Could we pass the method directly as a parameter instead of a `HookType` and then call it? Then we wouldn't need to do a switch or have a `HookType` object.

src/main/java/bio/terra/stairway/Stairway.java
```diff
@@ -189,6 +201,37 @@ public Stairway(Stairway.Builder builder) {
...
+      }
+
+      @Override
+      public HookAction endStep(FlightContext context) {
+        return handleDeafultHook();
+      }
+    };
+  }
+
+  private HookAction handleDeafultHook() {
```
fboulnois, line 232: minor fix:

```suggestion
  private HookAction handleDefaultHook() {
```

src/main/java/bio/terra/stairway/HookWrapper.java
fboulnois: We might want to have a separate return, like `IS_NULL`, or if we use `CONTINUE` then set a boolean that can be used to signal that the `null` condition was reached.

src/main/java/bio/terra/stairway/Stairway.java
fboulnois: Based on other code I looked at, this seems fine.

src/main/java/bio/terra/stairway/HookWrapper.java
```diff
@@ -0,0 +1,70 @@
...
+public class HookWrapper {
+  private static final Logger logger = LoggerFactory.getLogger(HookWrapper.class);
+  private Stairway stairway;
+
+  public HookWrapper(Stairway stairway) {
+    this.stairway = stairway;
+  }
+
+  public HookAction startFlight(FlightContext context) {
+    return handleHook(context, HookType.startFlight);
```
ddietterich, line 15: I'm not a fan of this pattern the takes the specific endpoint, makes it general, only to switch to make it specific again.
I'd probably have just cut'n'pasted. The cool kids pass functions as parameters. That might work here.

src/main/java/bio/terra/stairway/HookWrapper.java
```diff
@@ -0,0 +1,70 @@
+package bio.terra.stairway;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class HookWrapper {
```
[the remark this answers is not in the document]
snf2ye: @ddietterich I think I addressed this with my updates, but worth a double check.

src/main/java/bio/terra/stairway/HookWrapper.java
snf2ye: With the other changes, I am only passing through the StairwayHooks object, so this shouldn't be an issue anymore!

src/main/java/bio/terra/stairway/HookWrapper.java
snf2ye: I opted to just go the simpler route and have a bit of repetitive code, but thanks - good idea.

src/main/java/bio/terra/stairway/Stairway.java
snf2ye: No longer needed, but thanks 😞

src/main/java/bio/terra/stairway/Stairway.java
```diff
@@ -189,6 +201,37 @@ public Stairway(Stairway.Builder builder) {
     this.projectId = builder.projectId;
     this.workQueueEnabled = (projectId != null);
     this.quietingDown = new AtomicBoolean();
+    this.stairwayHook =
+        (builder.stairwayHook == null) ? populateDefaultStairwayHook() : builder.stairwayHook;
```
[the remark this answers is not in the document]
snf2ye: This is much better. Thank you.

src/main/java/bio/terra/stairway/HookWrapper.java
snf2ye: I opted for the cut'n'pasted. I was attempting to avoid duplicated code, but this is also a fair point!

src/main/java/bio/terra/stairway/HookWrapper.java
ddietterich: I don't know what IS_NULL info would be used for. I would leave it as is and plumb in an alternate return if/when that comes up.

src/main/java/bio/terra/stairway/HookWrapper.java
```diff
@@ -0,0 +1,77 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(HookWrapper.class);
+  private StairwayHook stairwayHook;
+
+  public HookWrapper(StairwayHook stairwayHook) {
+    this.stairwayHook = stairwayHook;
+  }
+
+  public HookAction startFlight(FlightContext context) {
+    if (stairwayHook == null) {
```
fboulnois, line 15: Since there is a significant amount of duplicated code between each HookWrapper method and it's more than 1-2 lines, I think it might be worthwhile having a private method that takes the right StairwayHook method instead.

ddietterich: That was the original approach that @snf2ye had coded. I pushed back, because the bulk of the private method was a switch statement to decide which hook to call.

The other alternative would be a private method that took the function to call as a parameter. That adds a lot of mechanism to save an if statement and a try-block in my view.

This may be one of those things where there is no way to make it beautiful code.

PR #16, opened 2020-06-04
DR-835 stairway orphans

ddietterich:
This is a large PR that restructures Stairway to support a cluster configuration and recover from failures of Stairway instances in that configuration. Major changes:
1. Make migrate work like TDRs. More cleanly driven from JobService.
2. Cover all DAO update operations with serialization retry
3. Recovery is entirely rewritten to support the different cases and new states
4. Admission control - tracking active work and limiting how many flights are queued locally, vs in the PubSub queue.

There are two areas that need design and (re)implementation ITFOT:
* Handling detection of failures of background threads
* Contention in the recovery process can lead to many duplicate entries in the work queue. For a small number of instances it is not a big deal, but when more than a handful, it could be an issue.


src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -159,6 +150,35 @@ private String lookupStairwayInstanceQuery(Connection connection, String stairwa
...
+  List<String> getStairwayInstanceList() throws DatabaseOperationException {
+    final String sql = "SELECT stairway_name FROM " + STAIRWAY_INSTANCE_TABLE;
+    List<String> instanceList = new ArrayList<>();
+
+    try (Connection connection = dataSource.getConnection()) {
+      startReadOnlyTransaction(connection);
+
+      try (NamedParameterPreparedStatement instanceStatement =
+              new NamedParameterPreparedStatement(connection, sql);
+          ResultSet rs = instanceStatement.getPreparedStatement().executeQuery()) {
```
fboulnois, line 169: minor nit: weird indentation here?

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -314,19 +338,99 @@ private void updateFlightState(String comment, String sql, FlightContext flightC
...
+   * obsolete stairway instance from the stairway_instance table.
+   *
+   * @param stairwayId the id of the, presumably deleted, stairway instance
+   * @return list of the flight ids that we disowned.
+   * @throws DatabaseOperationException on database error
+   * @throws InterruptedException interruption of the retry loop
+   */
+  void disownRecovery(String stairwayId) throws DatabaseOperationException, InterruptedException {
+    final String sql =
+        "UPDATE "
```
fboulnois, line 368: Do any of these need `@Transactional`-type annotations?

src/main/java/bio/terra/stairway/Queue.java
```diff
@@ -62,47 +67,65 @@
   private final Stairway stairway;
   private final Publisher publisher;
   private final String subscriptionName;
-  private SubscriberStub subscriber;
+  private SubscriberStub subscriberStub;
 
-  public Queue(Stairway stairway, String projectId, String subscriptionId, String topicId)
+  public Queue(Stairway stairway, String projectId, String topicId, String subscriptionId)
```
fboulnois, line 72: Any particular reason for the reorder? Is it due to the hierarchy of project -> topic -> subscription?

src/test/java/bio/terra/stairway/fixtures/TestUtil.java
```diff
@@ -0,0 +1,158 @@
...
+    return makeStairway(stairwayName, !continuing, !continuing, null);
+  }
+
+  public static Stairway setupConnectedStairway(String stairwayName, boolean continuing)
+      throws Exception {
+    String projectId = getEnvVar("GOOGLE_CLOUD_PROJECT", null);
+    if (projectId == null) {
+      throw new IllegalStateException(
+          "You must have GOOGLE_CLOUD_PROJECT and "
+              + "GOOGLE_APPLICATION_CREDENTIALS envvars defined");
```
fboulnois, line 76: Does this need to point to the `.pem` file?

README.md
```diff
@@ -96,17 +96,58 @@ Terra Data Repository (Jade) uses Stairway and has many examples of its use.
...
+result of a pod list with the recorded instance list returned from initialization.  
+
+Stairway recover-and-start performs recovery on the obsolete instances. It then records this new instance
+and starts accepting flights from the Stairway API and from the Work Queue, if enabled.
 
+### Execution Model
 Stairway provides a set of library calls for flight submission, status check, and waiting for completion.
 These calls all run on the calling thread.
 
+LIES and MORE LIES - this needs to be rewritten
```
fboulnois, line 149: 👍

src/main/java/bio/terra/stairway/FlightDao.java
marikomedlock: I think no, because he's doing "manual" transaction start/end in the method.

src/main/java/bio/terra/stairway/Queue.java
```diff
@@ -62,47 +67,65 @@
...
+    logger.info("Construct topic name");
     TopicName topicName = TopicName.ofProjectTopicName(projectId, topicId);
 
-    try (TopicAdminClient topicAdminClient = TopicAdminClient.create()) {
+    logger.info("Construct credentials");
+    GoogleCredentialsProvider credentialsProvider =
+        GoogleCredentialsProvider.newBuilder()
+            .setScopesToApply(Collections.singletonList("https://www.googleapis.com/auth/pubsub"))
+            .build();
+    logger.info("Credentials are: " + credentialsProvider);
```
marikomedlock, line 112: Did you mean to log credentialsProvider.getCredentials() or some other property instead of the object? And regardless, is this okay security-wise?

src/test/java/bio/terra/stairway/fixtures/TestUtil.java
marikomedlock, line 76: I think this is the same as the helper method getProjectId below.

src/main/java/bio/terra/stairway/FlightDao.java
ddietterich: This project uses the google java formatting and the idea plugin for same. I've noticed a few oddities.

I reformatted, but the formatter may switch it back: it auto-formats on build.

src/main/java/bio/terra/stairway/Queue.java
ddietterich: If I  recall correctly, i had used different orders for different methods. this order made more sense to me because it is in the right hierarchy.

src/test/java/bio/terra/stairway/fixtures/TestUtil.java
ddietterich: the application credentials point to third json file containing the credentials, that was pulled down from the cloud console when the service account key was created.

src/main/java/bio/terra/stairway/Queue.java
ddietterich: Good catch. I will remove it.

src/test/java/bio/terra/stairway/fixtures/TestUtil.java
ddietterich: I will use the helper method.

PR #17, opened 2020-07-08
DR-1124 Allow the work queue to be supplied by the caller

ddietterich:
Added parameters to specify the topic and subscription. Fix connected tests and added a new test the new feature.

### DIFF
--- a/src/main/java/bio/terra/stairway/QueueCreate.java
+++ b/src/main/java/bio/terra/stairway/QueueCreate.java
@@ -1,0 +1,111 @@
+package bio.terra.stairway;
+
+// In most cases we expect callers to provide case subscription and a topic. For backward
+// compatibility and
+// testing situation stairways able to create subscriptions and topics itself. This module
+// encapsulates
+// the creation operations.
+
+import com.google.api.gax.core.GoogleCredentialsProvider;
+import com.google.api.gax.rpc.AlreadyExistsException;
+import com.google.cloud.pubsub.v1.SubscriptionAdminClient;
+import com.google.cloud.pubsub.v1.SubscriptionAdminSettings;
+import com.google.cloud.pubsub.v1.TopicAdminClient;
+import com.google.cloud.pubsub.v1.TopicAdminSettings;
+import com.google.protobuf.Duration;
+import com.google.pubsub.v1.ProjectSubscriptionName;
+import com.google.pubsub.v1.Subscription;
+import com.google.pubsub.v1.TopicName;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.concurrent.TimeUnit;
+
+@SuppressFBWarnings(
+    value = "RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE",
+    justification = "Spotbugs doesn't understand resource try construct")
+public class QueueCreate {
+  private static final Logger logger = LoggerFactory.getLogger(QueueCreate.class);
+
+  // -- Queue Parameters --
+  // These parameters are used when constructing the queue and the subscriber. I don't think they
+  // need to be settable, but we may want to tune them as we get experience with the queue.
+  private static final int MAX_INBOUND_MESSAGE_BYTES = 10000;
+  private static final long MESSAGE_RETENTION_SECONDS = TimeUnit.DAYS.toSeconds(3);
+  private static final int ACK_DEADLINE_SECONDS = 100;
+
+  public static void makeTopic(String projectId, String topicId) throws IOException {
+    TopicName topicName = TopicName.ofProjectTopicName(projectId, topicId);
+
+    logger.debug("Construct credentials");
+    GoogleCredentialsProvider credentialsProvider =
+        GoogleCredentialsProvider.newBuilder()
+            .setScopesToApply(Collections.singletonList("https://www.googleapis.com/auth/pubsub"))
+            .build();
+    TopicAdminSettings topicAdminSettings =
+        TopicAdminSettings.newBuilder().setCredentialsProvider(credentialsProvider).build();
+
+    logger.debug("Try to create the topic");
+    try (TopicAdminClient topicAdminClient = TopicAdminClient.create(topicAdminSettings)) {
+      topicAdminClient.createTopic(topicName);
+      logger.debug("Created topic: " + topicId);
+    } catch (AlreadyExistsException ex) {
+      logger.debug("Topic already exists: " + topicId);
+    }
+  }
+
+  public static void makeSubscription(String projectId, String topicId, String subscriptionId)
+      throws IOException {
+    logger.debug("Construct names");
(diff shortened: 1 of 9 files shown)


src/test/java/bio/terra/stairway/QueueFlightTest.java
```diff
@@ -130,4 +132,90 @@ public void admissionControlTest() throws Exception {
     assertThat(
         "1st flight succeeded", flightState.getFlightStatus(), equalTo(FlightStatus.SUCCESS));
   }
+
+  @Test
+  public void supplyQueueTest() throws Exception {
```
snf2ye, line 137: Is there anything that should be cleaned up after this test?

src/main/java/bio/terra/stairway/Stairway.java
```diff
@@ -258,20 +279,37 @@ public Stairway(Stairway.Builder builder) {
...
 
+    this.workQueueProjectId = builder.workQueueProjectId;
+    this.workQueueEnabled = (workQueueProjectId != null && builder.enableWorkQueue);
+    if (!workQueueEnabled) {
+      workQueueCreate = false;
+      this.workQueueTopicId = null;
+      this.workQueueSubscriptionId = null;
+    } else {
+      if (builder.workQueueTopicId == null) {
+        workQueueCreate = true;
```
snf2ye, line 298: Just sanity checking that this is the desired behavior: Your above comment [line 194] says that if the workQueueProjectId is not supplied, no queue will be used. This set of logics appears to allow enableWorkQueue override this - if enableWorkQueue is true, then workQueueProjectId could be null and we would still use the work queue.

src/test/java/bio/terra/stairway/QueueTest.java
```diff
@@ -19,14 +20,22 @@
...
+    QueueCreate.makeTopic(projectId, topicId);
+    QueueCreate.makeSubscription(projectId, topicId, subscriptionId);
+    workQueue = new Queue(null, projectId, topicId, subscriptionId);
     workQueue.purgeQueue();
   }
 
+  @AfterEach
+  public void teardown() throws Exception {
+    //    QueueCreate.deleteQueue(projectId, topicId, subscriptionId);
+  }
```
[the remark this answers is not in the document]
ddietterich: I rearrange the code to remove the delete test and always do it and tear down.

src/main/java/bio/terra/stairway/QueueCreate.java
```diff
@@ -0,0 +1,111 @@
...
+    logger.debug("Try to create the topic");
+    try (TopicAdminClient topicAdminClient = TopicAdminClient.create(topicAdminSettings)) {
+      topicAdminClient.createTopic(topicName);
+      logger.debug("Created topic: " + topicId);
+    } catch (AlreadyExistsException ex) {
+      logger.debug("Topic already exists: " + topicId);
+    }
+  }
+
+  public static void makeSubscription(String projectId, String topicId, String subscriptionId)
```
fboulnois, line 60: In this implementation, we have a 1-1 mapping between topics and subscriptions, correct?

src/test/java/bio/terra/stairway/QueueFlightTest.java
ddietterich: I don't think so. We could clean up the queue, but that it is happily reused just like in the product. We test creation and deletion in the other queue test.

src/main/java/bio/terra/stairway/Stairway.java
ddietterich: `this.workQeueEnabled` is only true if `enableWorkQueue` is true and `builder.workQueueProjectId` is not null.
I think this is ok

src/main/java/bio/terra/stairway/Stairway.java
```diff
@@ -258,20 +279,37 @@ public Stairway(Stairway.Builder builder) {
...
+        this.workQueueSubscriptionId = stairwayClusterName + "-workqueue-sub";
+      } else {
+        workQueueCreate = false;
+        this.workQueueTopicId = builder.workQueueTopicId;
+        this.workQueueSubscriptionId = builder.workQueueSubscriptionId;
+        if (workQueueSubscriptionId == null) {
+          throw new StairwayExecutionException(
+              "subscription id must be specified if topic id is specified");
+        }
+      }
```
fboulnois, line 309: I wanted to check if this reformulation made the code clearer in my mind. I'm not sure it does, although it's a bit more concise. Feel free to ignore.

```suggestion
      workQueueCreate = builder.workQueueTopicId == null;
      this.workQueueTopicId = workQueueCreate ? stairwayClusterName + "-workqueue" : builder.workQueueTopicId;
      this.workQueueSubscriptionId = workQueueCreate ? stairwayClusterName + "-workqueue-sub" : builder.workQueueSubscriptionId;
       if (workQueueSubscriptionId == null) {
          throw new StairwayExecutionException(
              "subscription id must be specified if topic id is specified");
       }
```

src/main/java/bio/terra/stairway/QueueCreate.java
```diff
@@ -0,0 +1,111 @@
...
+  private static final long MESSAGE_RETENTION_SECONDS = TimeUnit.DAYS.toSeconds(3);
+  private static final int ACK_DEADLINE_SECONDS = 100;
+
+  public static void makeTopic(String projectId, String topicId) throws IOException {
+    TopicName topicName = TopicName.ofProjectTopicName(projectId, topicId);
+
+    logger.debug("Construct credentials");
+    GoogleCredentialsProvider credentialsProvider =
+        GoogleCredentialsProvider.newBuilder()
+            .setScopesToApply(Collections.singletonList("https://www.googleapis.com/auth/pubsub"))
```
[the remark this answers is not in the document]
ddietterich: I thought you had to. I was following documentation example.

src/main/java/bio/terra/stairway/QueueCreate.java
ddietterich: Correct. In our test environment i expect we will continue to have a 1-1 mapping. Now that stairway will accept the topic and subscription, the the caller can organize the topic and subscription anyway they please.

src/main/java/bio/terra/stairway/Stairway.java
ddietterich: The reason i did it the way i did it is to allow the TopicId and SubscriptionId to be final.

src/main/java/bio/terra/stairway/QueueCreate.java
fboulnois: I like this design since it simplifies things considerably.

src/main/java/bio/terra/stairway/QueueCreate.java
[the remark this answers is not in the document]
fboulnois: I think if you create a list with just "pubsub", it should work. We do something similar in the [data-repo-client-example](https://github.com/broadinstitute/jade-data-repo-client-example/blob/e024d443ca78d2d5743db15a9064318bf4cfda19/src/main/java/DataRepoClientExample.java#L14). I don't have a strong inclination either way, although one issue I can think of is Google retiring or redirecting some API endpoints in the future.

PR #20, opened 2020-07-15
DR-1154 fix recovery bug

ddietterich:


src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -418,13 +419,14 @@ void disownRecovery(String stairwayId) throws DatabaseOperationException, Interr
         NamedParameterPreparedStatement statement =
             new NamedParameterPreparedStatement(connection, sql)) {
 
-      startReadOnlyTransaction(connection);
+      startTransaction(connection);
```
fboulnois, line 422: Would you kindly comment on why this was necessary?

src/main/java/bio/terra/stairway/Flight.java
```diff
@@ -154,6 +155,10 @@ private FlightStatus fly() throws InterruptedException {
       // Dismal failure - undo failed!
       context().setResult(undoResult);
 
+    } catch (InterruptedException ex) {
+      // Interrupted exception - we assume this means that the thread pool is shutting down and
+      // forcibly stopping all threads. We propagate the exception.
+      throw ex;
```
snf2ye, line 161: If I'm understanding correctly, by throwing the error, we're saying that this failed for extraneous reasons, so it was not a step failure.
Just curious - What will happen when this flight is picked up again? Will it just retry this step? If that's the case, why not set the step status to STEP_RESULT_FAILURE_RETRY?

src/main/java/bio/terra/stairway/FlightDao.java
ddietterich: My concern, may be unfounded, was that the snapshot transaction would be reading out of date data because there would be too many other transactions writing to the table. This change makes sure that the read transaction is ordered with the write transactions. I will add a comment.

src/main/java/bio/terra/stairway/Flight.java
ddietterich: Not exactly. The InterruptedException is used in the specific case where the thread pool is terminating all of the flight threads during pod termination. The exception is caught two frames above and handled there. About line 95.

The result is that the flight is disowned - this stairway instance cannot own it since it is terminating - and will be picked up by another stairway instance. It will start at the step that was interrupted. That was the bug before: the fight with restarting at the next step.

By design, all steps have to be idempotent.

PR #37, opened 2020-08-31
DR-1297 Provide step class names for logging; quiet down queue logging

ddietterich:
There are three changes:
1. Set the scheduling and queue logging to debug. get rid of that every five second blat in our data repo logs
2. Build a list of step class names as the flight is being constructed. pass those names over to the flight context so they are available to the hooks.
3. Opportunistically add `final` where IntelliJ tells me to

### DIFF
--- a/src/main/java/bio/terra/stairway/WorkQueueListener.java
+++ b/src/main/java/bio/terra/stairway/WorkQueueListener.java
@@ -28,7 +28,7 @@ public class WorkQueueListener implements Runnable {
   // happens we sleep for NO_PULL_SLEEP_SECONDS and then check the queue depth.
   //
   // The reason to limit to MAX_MESSAGES_PER_PULL is that the evaluation of queue depth is done at,
-  // say,time T0, but the pull will wait for message arrival from pubsub at, say, time T1.
+  // say, time T0, but the pull will wait for message arrival from pubsub at, say, time T1.
   // By time T1, the queue depth situation might have changed. So we don't want to set up to grab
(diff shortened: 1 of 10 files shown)


src/main/java/bio/terra/stairway/WorkQueueListener.java
```diff
@@ -24,22 +25,18 @@ public WorkQueueListener(Stairway stairway, Queue workQueue) {
...
+  // happens we sleep for NO_PULL_SLEEP_SECONDS and then check the queue depth.
   //
   // The reason to limit to MAX_MESSAGES_PER_PULL is that the evaluation of queue depth is done at,
-  // say,
-  // time T0, but the pull will wait for message arrival from pubsub at, say, time T1. By time T1,
-  // the
-  // queue depth situation might have changed. So we don't want to set up to grab lots of messages
-  // and
-  // then find we are queuing flights way over maxQueuedFlights.
+  // say,time T0, but the pull will wait for message arrival from pubsub at, say, time T1.
```
fboulnois:
```suggestion
  // say, time T0, but the pull will wait for message arrival from pubsub at, say, time T1.
```

snf2ye (approved):
Thank you - I think these logs will be helpful!

fboulnois (approved):
lgtm, micro-nit 🦠 :

src/main/java/bio/terra/stairway/WorkQueueListener.java
ddietterich: Thanks for that!

PR #39, opened 2020-10-13
DR-1408 Run hookWrapper#endStep even when the step throws an exception

wchamber:
- Put hookWrapper().endStep in a finally so that it is always run, even if the step ends by throwing an exception.
- Fix bash script for setting variables by removing whitespace.

Right now, if the step returns a StepResult error, the endStep on a StairwayHook will run. If the step throws an exception however, Stairway will catch the exception and wrap it as a StepResult error, but not run endStep. We should run the StairwayHook#endStep regardless of how the step finishes.

### DIFF
--- a/src/main/java/bio/terra/stairway/Flight.java
+++ b/src/main/java/bio/terra/stairway/Flight.java
@@ -253,7 +253,6 @@ public class Flight implements Runnable {
         } else {
           result = currentStep.step.undoStep(context());
         }
-        hookWrapper().endStep(flightContext);
       } catch (InterruptedException ex) {
         // Interrupted exception - we assume this means that the thread pool is shutting down and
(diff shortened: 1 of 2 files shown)


ddietterich (commented):
Thanks for the code! This needs a Jira ticket before merge. We need to be able to correlate every merge with a ticket. I can make one for you if you like.

wchamber:
> Thanks for the code! This needs a Jira ticket before merge. We need to be able to correlate every merge with a ticket. I can make one for you if you like.

Created https://broadworkbench.atlassian.net/browse/DR-1408. Let me know if I've done it wrong!

ddietterich (approved):
lgtm

ddietterich:
All set. Thank you!

PR #40, opened 2020-10-21
DR-1444: Truncate FlightMap toString

snf2ye:
Logging of input parameters to the snapshot create flight is flooding out connected test logs. Example [logs here](https://github.com/DataBiosphere/jade-data-repo/runs/1283857609?check_suite_focus=true#step:5:33834). 
Solution: FlightMap.toString() now truncates any value at 500 characters.

### DIFF
--- a/src/main/java/bio/terra/stairway/FlightContext.java
+++ b/src/main/java/bio/terra/stairway/FlightContext.java
@@ -194,19 +194,4 @@ public class FlightContext {
         .append("flightStatus", flightStatus)
         .toString();
   }
-
-  public String toStringTrunc() {
-    return new ToStringBuilder(this)
-        .append("stairway", stairway)
-        .append("flightId", flightId)
-        .append("flightClassName", flightClassName)
-        // .append("inputParameters", inputParameters)
-        .append("workingMap", workingMap)
-        .append("stepIndex", stepIndex)
-        .append("rerun", rerun)
-        .append("direction", direction)
-        .append("result", result)
-        .append("flightStatus", flightStatus)
-        .toString();
-  }
 }
--- a/src/main/java/bio/terra/stairway/FlightContext.java
+++ b/src/main/java/bio/terra/stairway/FlightContext.java
@@ -194,4 +194,19 @@ public class FlightContext {
         .append("flightStatus", flightStatus)
         .toString();
   }
+
+  public String toStringTrunc() {
+    return new ToStringBuilder(this)
+        .append("stairway", stairway)
+        .append("flightId", flightId)
+        .append("flightClassName", flightClassName)
+        // .append("inputParameters", inputParameters)
+        .append("workingMap", workingMap)
+        .append("stepIndex", stepIndex)
+        .append("rerun", rerun)
+        .append("direction", direction)
+        .append("result", result)
+        .append("flightStatus", flightStatus)
+        .toString();
+  }
 }
--- a/src/main/java/bio/terra/stairway/FlightMap.java
+++ b/src/main/java/bio/terra/stairway/FlightMap.java
@@ -11,8 +11,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import org.apache.commons.lang3.builder.ToStringBuilder;
-import org.apache.commons.lang3.builder.ToStringStyle;
 
 /**
  * FlightMap wraps a {@code HashMap<String, Object>} It provides a subset of the HashMap methods. It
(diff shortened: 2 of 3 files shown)


src/main/java/bio/terra/stairway/FlightContext.java
```diff
@@ -194,4 +194,19 @@ public String toString() {
         .append("flightStatus", flightStatus)
         .toString();
   }
+
+  public String toStringTrunc() {
+    return new ToStringBuilder(this)
+        .append("stairway", stairway)
+        .append("flightId", flightId)
+        .append("flightClassName", flightClassName)
+        // .append("inputParameters", inputParameters)
```
nmalfroy: got an extra comment here

src/main/java/bio/terra/stairway/FlightContext.java
```diff
@@ -194,4 +194,19 @@ public String toString() {
...
+        .append("workingMap", workingMap)
```
nmalfroy: Do you think it would make sense to show input parameters up to a point?  Say, show the first 500 characters or something like that?  For cases where the input is pretty small, it still might be helpful to have this information.  Do the inputs get stored in the DB?  I just want to make sure that we don't lose information that may end up being critical to debug.

nmalfroy (commented):
Is this what was causing that massive log message?  Just wondering if we want to attempt to show some of the information, not just hide the field outright?

src/main/java/bio/terra/stairway/FlightContext.java
```diff
@@ -194,4 +194,19 @@ public String toString() {
         .append("flightStatus", flightStatus)
         .toString();
   }
+
+  public String toStringTrunc() {
```
fboulnois: A comment here indicating which parameter is missing would be useful. If you are not planning to truncate any strings (as suggested by @nmalfroy below), then I would rename the function to `toStringQuiet` or `toStringTerse` or something along those lines.

fboulnois (commented):
A suggestion:

src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -110,8 +108,23 @@ public void fromJson(String json) {
...
-    return new ToStringBuilder(this, ToStringStyle.JSON_STYLE).append("map", map).toString();
+    int truncateLength = 500;
+    StringBuilder sb = new StringBuilder("{");
+    map.forEach(
+        (k, v) -> {
+          int valLength = v.toString().length();
+          sb.append(
+              k
+                  + "="
+                  + v.toString()
```
pshapiro4broad: This computes `v.toString()` twice which, if it expands to a lot of text, can be expensive. It might be better to do it once before line 118 instead.

src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -110,8 +108,23 @@ public void fromJson(String json) {
...
+                      .substring(0, valLength > truncateLength ? truncateLength : valLength)
```
pshapiro4broad: This is `min`
```suggestion
                      .substring(0, Math.min(valLength, truncateLength))
```

src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -110,8 +108,23 @@ public void fromJson(String json) {
...
+                  + ", ");
+        });
+    sb.delete(sb.length() - 2, sb.length());
```
pshapiro4broad: It would be nice to not have to do it this way. Also, what if `map` is empty? I think you'd get an exception.

IMO using a stream over the map would be better. Something like
```
map.entrySet().stream().map(...).collect(Collectors.joining(", "))
```

pshapiro4broad (commented):
A few minor comments

src/main/java/bio/terra/stairway/FlightMap.java
snf2ye: So much better. Thank you!

src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -110,8 +109,23 @@ public void fromJson(String json) {
...
   public String toString() {
-    return new ToStringBuilder(this, ToStringStyle.JSON_STYLE).append("map", map).toString();
+    int truncateLength = 500;
+    StringBuilder sb = new StringBuilder("{");
+    sb.append(
+        map.entrySet().stream()
+            .map(
+                set -> {
+                  String valString = set.getValue().toString();
+                  return String.join(
```
pshapiro4broad: It wouldn't have occurred to me to use `String.join()` to join two things, is there a reason to use this over `+`? Either way is OK with me.

src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -110,8 +109,23 @@ public void fromJson(String json) {
...
+  // Truncate working map to only print first 500 characters
   @Override
   public String toString() {
-    return new ToStringBuilder(this, ToStringStyle.JSON_STYLE).append("map", map).toString();
+    int truncateLength = 500;
+    StringBuilder sb = new StringBuilder("{");
+    sb.append(
+        map.entrySet().stream()
+            .map(
+                set -> {
```
pshapiro4broad: This is such a minor thing, but I would call this `entry` not `set` since it's an entry in the Map (and its type is `Map.Entry`).

pshapiro4broad (approved):
Looks good

src/main/java/bio/terra/stairway/FlightMap.java
snf2ye: Is it probably overkill. I don't really have a preference - I'll switch it back.

ddietterich (approved):
lgtm

PR #41, opened 2020-10-26
DR-1444 Follow up: Flight map to string fix

snf2ye:


RoriCremer (approved):
so fast!

pshapiro4broad (approved):

src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -118,7 +118,7 @@ public String toString() {
         map.entrySet().stream()
             .map(
                 entry -> {
-                  String valString = entry.getValue().toString();
+                  String valString = String.valueOf(entry.getValue());
```
nmalfroy, line 121: will that return a null if entry.getValue() is null?

pshapiro4broad: It will return `"null"`

nmalfroy: Ah got it.  thanks

snf2ye: I don't think we actually want to add this test, but this shows the behavior:
https://github.com/DataBiosphere/jade-data-repo/compare/sh-add-string-test?expand=1

With this change, the output just add "null" as a string for the value.
{key=null}

PR #42, opened 2020-11-16
Break duplicate flightID into its own exception

zloery:
This adds `DuplicateFlightIdException` as a checked exception in the FlightDAO in the case where a flight is submitted with an ID that's already in use. This is useful for clients providing a flight ID to distinguish duplicate ID exceptions from other generic DB errors.


ddietterich (approved):
lgtm

PR #44, opened 2020-12-02
[PF-242] Add StepHook in StairwayHook to support sharing objects between startStep() and endStep()  

yonghaoy:
Reason we want that: In Buffer Service, we want to use Hook to track the tracing of a flight. And we will have something like this:
```
startStep(FlightContext context) {
  startSpan()
}

endStep(FlightContext context) {
 endSpan()
}
```

But to end the span. we need to get the spanContext which is created in startSpan(). 

Would be great if we can pass the spanContext around.


wchamber (commented):
I was thinking we would have a little more here for a draft suggestion, like the modifications to the StairwayHook interface as well as Flight.java/HookWrapper.java (if not totally working, enough to see the use). But we can start with this if you like.

PR #45, opened 2020-12-04
TN PF-261 Initial implementation of FlightDebugInfo

tnadler:
This Pull Request has the initial implementation for flight specific debug information and the first field of that object.

FlightDebugInfo is submitted with the initial flight description. RestartEachStep causes it to pause after each step to be restarted by the Stairway. This ensures the flight can be properly recreated from memory.


tnadler:
Dan had good comments on flight submission. Reworking so closing for now.

PR #47, opened 2020-12-08
TN PF-261 Flight Debug Info

tnadler:
This PR contains the changes to:
1) Create a FlightDebugInfo class and pass it to the flight constructor.
2) Store that class as a JSON in the flight table (includes liquibase config to add a column to the table).
3) Logic to implement RestartEachStep - a way of running a flight where the flight stops and is resumed from the database at each step.

### DIFF
--- a/src/main/java/bio/terra/stairway/FlightDebugInfo.java
+++ b/src/main/java/bio/terra/stairway/FlightDebugInfo.java
@@ -2,22 +2,31 @@ package bio.terra.stairway;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.Map;
 
 /**
  * Debug information for a flight. Parameters here change how flights run to ensure debugability/
- * testability. Designed to be easily extensible,
+ * testability.
  */
 public class FlightDebugInfo {
 
   private static ObjectMapper objectMapper;
-  private boolean restartEachStep; // true - restart the flight at each step
+  private boolean restartEachStep; // if true - restart the flight at each step
+  private Map<Integer, StepStatus> failAtSteps; // For each entry, the flight will return a given
+  // StepStatus for the step at the given position.
 
   // Use a builder so it is easy to add new fields
   public static class Builder {
     private boolean restartEachStep;
+    private Map<Integer, StepStatus> failAtSteps;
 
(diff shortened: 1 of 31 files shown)


src/main/java/bio/terra/stairway/FlightDebugInfo.java
```diff
@@ -0,0 +1,65 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * Debug information for a flight. Parameters here change how flights run to ensure debugability/
+ * testability. Designed to be easily extensible,
+ */
+public class FlightDebugInfo {
+
+  private static ObjectMapper objectMapper;
```
ddietterich: these can be `final` I think

src/test/java/bio/terra/stairway/DebugInfoTest.java
```diff
@@ -0,0 +1,88 @@
...
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The debug info test needs to ensure that the flight actually pauses between steps so it can be
+ * restarted. To do this we use a two step Flight. We start the flight, and then ensure after the
+ * initial submission the flight is left in a ready state and can be resumed. NOTE: This only works
+ * for the initial submit currently as we have not persisted FlightDebugInfo for reading from the
```
ddietterich: Well, now you have! But the flight won't resume yet.

ddietterich (commented):
A few nit comments. Otherwise, looks good, except for the constructor thing...

src/test/java/bio/terra/stairway/DebugInfoTest.java
tnadler: This file did not get updated somehow! Replaced all this text.!

src/main/java/bio/terra/stairway/FlightDebugInfo.java
tnadler: Initializing object mapper this way looks odd to me somehow...
The restartEachStep can't (or at least I think can't) as to get Jackson to work I created a setter.

ddietterich (approved):
LGTM - thank you!

PR #48, opened 2020-12-11
Add support for FailAtSteps to FlightDebugInfo. This mode allows user…

tnadler:
…s to insert failures at specific indices of the flight.

### DIFF
--- a/src/main/java/bio/terra/stairway/Flight.java
+++ b/src/main/java/bio/terra/stairway/Flight.java
@@ -7,8 +7,10 @@ import static bio.terra.stairway.FlightStatus.WAITING;
 import bio.terra.stairway.exception.DatabaseOperationException;
 import bio.terra.stairway.exception.RetryException;
 import bio.terra.stairway.exception.StairwayExecutionException;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Set;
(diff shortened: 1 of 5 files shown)


src/main/java/bio/terra/stairway/Flight.java
```diff
@@ -263,8 +270,23 @@ private StepResult stepWithRetry() throws InterruptedException, StairwayExecutio
       try {
         // Do or undo based on direction we are headed
         hookWrapper().startStep(flightContext);
+
         if (context().isDoing()) {
           result = currentStep.step.doStep(context());
+
+          // If we are in debug mode and a failure is set for this step AND we have not already
+          // failed here, then insert a failure. We do this right after the step completes but
```
ddietterich, line 278: And the reason you only want to insert once is if there are retries?
I wonder if we should just allow that case: exhaustion of retries causing the transition to undo.
Or maybe extend the debug info to allow specifying the number of times we should insert the fault?

ddietterich (approved):
LGTM

I'd like to kick around thoughts about the fail-only-once semantic you've implemented, but I think those would be extensions to what you've got here.

src/main/java/bio/terra/stairway/Flight.java
tnadler: SG. As discussed offline, will augment if needed.

PR #49, opened 2020-12-23
[PF-316] Use DefaultTyping.EVERYTHING to let Jaskson embed final class

yonghaoy:
Seems EVERYTHING works for UUID, draft a PR with some sample tests. 
Not sure if that will break other parts. 
https://fasterxml.github.io/jackson-databind/javadoc/2.11/com/fasterxml/jackson/databind/ObjectMapper.DefaultTyping.html#EVERYTHING


ddietterich:
This technique is known not to work.

PR #50, opened 2021-01-05
Add log statement when flight is no longer retrying

tnadler:


src/main/java/bio/terra/stairway/Flight.java
```diff
@@ -340,6 +340,7 @@ private StepResult stepWithRetry() throws InterruptedException, StairwayExecutio
       }
     } while (currentStep.retryRule
         .retrySleep()); // retry rule decides if we should try again or not
+    logger.info("No longer retrying - retry rule has reached it's limit");
```
jaycarlton: *its. Maybe clarify if limit is a number of retries, wall clock time, or some other thing, and include values.

ddietterich: The reason for not retrying is embedded in the retry rule itself; we don't know that here. That would require logging within the specific retry rule. Also, it might be nice to skip logging in the `RetryRuleNone` case, where no retrying is expected.

Still, I'm OK with this change. What do you think @tovanadler ?

tnadler: Sure, happy to skip in that case. Basically was just adding a quick and dirty fix to address the fact that line 331 says we are retrying when we are not. WDYT now?

jaycarlton (approved):

src/main/java/bio/terra/stairway/Flight.java
jaycarlton: Maybe the retry rule should expose either a toString() or some failure status so logging can happen consistently? Or maybe all logging should be inside the rule.

ddietterich:
I made a ticket ([PF-351](https://broadworkbench.atlassian.net/browse/PF-351) for fixing the misleading logging.
I will post a PR for that - I hope by the end of the day.
@tnadler if it is OK with you, I'd like to close this PR.

tovanadler:
Sure, sg

PR #51, opened 2021-01-07
PF-351: Detailed retry logging

ddietterich:
Added detailed retry logging within the retry rules.
Fixed the wording of the outer retry log message so it is not misleading.
Added a test for randomBackoff.
Validated the new logging by examining the output of the retry test.

In addition, fixed a javadoc complaint (FlightFactory) and a spotbugs complaint (TestFlightMultiStepRetry)

### DIFF
--- a/src/main/java/bio/terra/stairway/RetryRuleExponentialBackoff.java
+++ b/src/main/java/bio/terra/stairway/RetryRuleExponentialBackoff.java
@@ -1,7 +1,7 @@
 package bio.terra.stairway;
 
 import java.time.Duration;
-import java.time.LocalDateTime;
+import java.time.Instant;
 import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -13,7 +13,7 @@ public class RetryRuleExponentialBackoff implements RetryRule {
   private final long maxIntervalSeconds;
   private final long maxOperationTimeSeconds;
 
-  private LocalDateTime endTime;
+  private Instant endTime;
   private long intervalSeconds;
 
   /**
@@ -33,12 +33,12 @@ public class RetryRuleExponentialBackoff implements RetryRule {
   @Override
   public void initialize() {
     intervalSeconds = initialIntervalSeconds;
-    endTime = LocalDateTime.now().plus(Duration.ofSeconds(maxOperationTimeSeconds));
+    endTime = Instant.now().plus(Duration.ofSeconds(maxOperationTimeSeconds));
   }
 
   @Override
   public boolean retrySleep() throws InterruptedException {
-    LocalDateTime now = LocalDateTime.now();
+    Instant now = Instant.now();
     if (now.isAfter(endTime)) {
       logger.info(
           "Retry rule exponential: now ({}) is past max time {} - not retrying", now, endTime);
--- a/src/main/java/bio/terra/stairway/RetryRuleExponentialBackoff.java
+++ b/src/main/java/bio/terra/stairway/RetryRuleExponentialBackoff.java
@@ -3,8 +3,12 @@ package bio.terra.stairway;
 import java.time.Duration;
 import java.time.LocalDateTime;
 import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class RetryRuleExponentialBackoff implements RetryRule {
+  private static final Logger logger = LoggerFactory.getLogger(RetryRule.class);
+
   private final long initialIntervalSeconds;
   private final long maxIntervalSeconds;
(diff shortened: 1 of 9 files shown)


src/main/java/bio/terra/stairway/RetryRuleExponentialBackoff.java
```diff
@@ -34,10 +38,19 @@ public void initialize() {
 
   @Override
   public boolean retrySleep() throws InterruptedException {
-    if (LocalDateTime.now().isAfter(endTime)) {
+    LocalDateTime now = LocalDateTime.now();
```
jaycarlton: Does this print with a time zone? Do we have a standard across logs for local times? I'm used to UTC.

src/main/java/bio/terra/stairway/RetryRuleFixedInterval.java
```diff
@@ -30,9 +34,16 @@ public void initialize() {
   @Override
   public boolean retrySleep() throws InterruptedException {
     if (retryCount >= maxCount) {
+      logger.info("Retry rule fixed: try {} of {} - not retrying", retryCount, maxCount);
```
jaycarlton: nit: So if there is a count of 3, does it show 0, 1, and 2 of 3? Maybe say something like "completed N attempts", if otherwise it's always 3/3. 4/3 would be a bug, if I'm u nderstanding.

src/main/java/bio/terra/stairway/RetryRuleFixedInterval.java
```diff
@@ -30,9 +34,16 @@ public void initialize() {
...
       return false;
     }
 
+    logger.info(
+        "Retry rule fixed: try {} of {} - retrying after sleep of {} seconds",
+        retryCount,
+        maxCount,
+        intervalSeconds);
```
jaycarlton, line 47: Can you use a Duration object for intervals that knows how to print itself? Maybe out of scope.

src/test/java/bio/terra/stairway/RetryTest.java
```diff
@@ -128,4 +128,22 @@ public void exponentialMaxIntervalSuccessTest() throws Exception {
...
+    inputParameters.put("retryType", "random");
+    inputParameters.put("failCount", 4);
+    inputParameters.put("operationMilliseconds", 500L);
+    inputParameters.put("maxConcurrency", 5);
+    inputParameters.put("maxCount", 3);
+
+    String flightId = "randomFailureTest";
+    stairway.submit(flightId, TestFlightRetry.class, inputParameters);
+
+    FlightState result = stairway.waitForFlight(flightId, null, null);
```
jaycarlton, line 145: Can we have a single-argument version of this method in Stairway? I don't like seeing nulls passed around explicitly.

src/test/java/bio/terra/stairway/flights/TestFlightRetry.java
```diff
@@ -29,11 +30,15 @@ public TestFlightRetry(FlightMap inputParameters, Object applicationContext) {
       retryRule =
           new RetryRuleExponentialBackoff(
               initialIntervalSeconds, maxIntervalSeconds, maxOperationTimeSeconds);
+    } else if (StringUtils.equals("random", retryType)) {
```
jaycarlton, line 33: Can we use an enum with string constant metadata for the retry type? E.g. `RETRY_TYPE.random.name().equals(retryType)`

jaycarlton (approved):
Just nits and questions. Thanks.

src/main/java/bio/terra/stairway/RetryRuleExponentialBackoff.java
ddietterich: Good catch! I have changed this to use `Instant` which provides a UTC time.

src/main/java/bio/terra/stairway/RetryRuleFixedInterval.java
ddietterich: If the maxCount 3, then it will retry 0, 1, and 2. When it hits 3, it stops. And 4/3 is a surprise.
Zero-based doesn't work with a "N of M" phrasing, does it. I'll make the counter 1-based.

src/test/java/bio/terra/stairway/RetryTest.java
ddietterich: That method is deprecated. I still use it in testing, but should factor it out into a test fixture. I will made a ticket for that: PF-353

src/test/java/bio/terra/stairway/flights/TestFlightRetry.java
ddietterich: It is testing code with one caller. I don't think it is worth it.

tnadler (approved):
Looks good to me, thank you!

PR #52, opened 2021-01-18
PF-311 hook enhancements

ddietterich:
The hook enhancements added in this PR are described here [Enhanced Hook Interface](https://docs.google.com/document/d/19CNrhoUbb5-vQFMyScgQTLwpMzl0GQDbiivAunxP18o/edit?pli=1&resourcekey=0-4JNqmHGPilkAyWAo2RmvIg).

Previously, Stairway minted its own unique ID for a Stairway instance. This makes logging in hooks problematic, because you cannot access the mapping from the ID to the Stairway instance. In this PR, I made the unique ID identical to the Stairway instance name. The instance name must already be unique, so there is no semantic difference. The result is that logging can include the Stairway instance name provided by the Stairway caller.

This PR also changes the logging on Stairway retries. Most logging is moved into the RetryRule subclasses and is made more informative. This addresses the confusing logging where Stairway logs "I'm retrying" and the RetryRule quietly says "no you are not".

I think most of the code is fairly obvious. One piece that deserves some explanation is the change to create a FlightContext during recovery. On recovery, we only update the flight row, so do not need the full FlightContext. The only reason to create the FlightContext is for state transition logging. Creating it costs an additional two queries in the database. My judgment was that getting detailed state transition logging was worth the cost. Further, the extra work is only during recovery - typically when a pod is starting up - and not during normal operation.

### DIFF
--- a/src/main/java/bio/terra/stairway/FlightContext.java
+++ b/src/main/java/bio/terra/stairway/FlightContext.java
@@ -1,8 +1,9 @@
 package bio.terra.stairway;
 
-import java.util.List;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.builder.ToStringBuilder;
+
+import java.util.List;
 
 /**
  * Context for a flight. This contains the full state for a flight. It is what is held in the
@@ -11,8 +12,7 @@ import org.apache.commons.lang3.builder.ToStringBuilder;
 public class FlightContext {
   private Stairway stairway; // the stairway instance running this flight
   private String flightId; // unique id for the flight
-  private final String
-      flightClassName; // class name of the flight;  for recreating the flight object
+  private final String flightClassName; // class name of the flight
   private final FlightMap inputParameters; // allows for reconstructing the flight; set unmodifiable
   private final FlightMap workingMap; // open-ended state used by the steps
   private int stepIndex; // what step we are on
@@ -22,6 +22,7 @@ public class FlightContext {
   private FlightStatus flightStatus;
   private List<String> stepClassNames;
   private FlightDebugInfo debugInfo;
+  private List<StepHook> stepHooks;
 
   // Construct the context with defaults
   public FlightContext(
@@ -133,6 +134,14 @@ public class FlightContext {
 
   public FlightDebugInfo getDebugInfo() {
     return debugInfo;
+  }
+
+  public List<StepHook> getStepHooks() {
+    return stepHooks;
+  }
+
+  public void setStepHooks(List<StepHook> stepHooks) {
+    this.stepHooks = stepHooks;
   }
 
   /**
--- a/src/main/java/bio/terra/stairway/FlightContext.java
+++ b/src/main/java/bio/terra/stairway/FlightContext.java
@@ -1,9 +1,8 @@
 package bio.terra.stairway;
 
+import java.util.List;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.builder.ToStringBuilder;
-
-import java.util.List;
 
 /**
  * Context for a flight. This contains the full state for a flight. It is what is held in the
--- a/src/main/java/bio/terra/stairway/FlightContext.java
+++ b/src/main/java/bio/terra/stairway/FlightContext.java
@@ -9,18 +9,56 @@ import org.apache.commons.lang3.builder.ToStringBuilder;
  * database for the flight and it is passed into the steps
  */
 public class FlightContext {
-  private Stairway stairway; // the stairway instance running this flight
-  private String flightId; // unique id for the flight
-  private final String flightClassName; // class name of the flight
-  private final FlightMap inputParameters; // allows for reconstructing the flight; set unmodifiable
-  private final FlightMap workingMap; // open-ended state used by the steps
-  private int stepIndex; // what step we are on
-  private boolean rerun; // true - rerun the current step
+  /** The stairway instance running this flight */
+  private Stairway stairway;
+
+  /**
+   * Identifier for the flight. The caller is expected to ensure that these are unique within this
+   * Stairway.
+   */
+  private String flightId;
+
+  /**
+   * Full class name of the flight. This is used to reconstruct the flight from its persisted state.
+   */
+  private final String flightClassName;
+
+  /** Unmodifiable flight map of the input parameters to the flight */
+  private final FlightMap inputParameters;
+
+  /**
+   * Modifiable flight map used to communicate state between steps and convey output of the flight.
+   */
+  private final FlightMap workingMap;
+
+  /** Index into the flight's step array of the step we are */
+  private int stepIndex;
+
+  /** Control flag to tell the flight runner to rerun the current step */
+  private boolean rerun;
+
+  /** Direction of execution: do, undo, switch. */
   private Direction direction;
-  private StepResult result; // current step status
+
+  /** Returned status of the current step */
+  private StepResult result;
+
+  /** State of this flight */
   private FlightStatus flightStatus;
+
+  /**
+   * List of the class names of the steps. This is not used by the flight execution code. It is here
+   * to allow more meaningful log messages by hooks.
+   */
   private List<String> stepClassNames;
+
+  /** Debug control of this flight. */
   private FlightDebugInfo debugInfo;
+
+  /**
+   * Dynamic list of step hooks defined for the current step, created using {@link
+   * StairwayHook#stepFactory(FlightContext)}
+   */
   private List<StepHook> stepHooks;
 
   // Construct the context with defaults
--- a/src/main/java/bio/terra/stairway/FlightDao.java
+++ b/src/main/java/bio/terra/stairway/FlightDao.java
@@ -370,7 +370,7 @@ class FlightDao {
           throw new DatabaseOperationException(
               String.format(
                   "Failed to update status to %s for flight: %s",
-                  flightContext.getFlightStatus().name(), flightContext.getFlightId()),
+                  flightContext.getFlightStatus(), flightContext.getFlightId()),
               ex);
         }
       }
--- a/src/main/java/bio/terra/stairway/FlightDao.java
+++ b/src/main/java/bio/terra/stairway/FlightDao.java
@@ -8,6 +8,10 @@ import bio.terra.stairway.exception.FlightFilterException;
 import bio.terra.stairway.exception.FlightNotFoundException;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.sql.DataSource;
 import java.sql.Connection;
 import java.sql.ResultSet;
 import java.sql.SQLException;
@@ -16,9 +20,6 @@ import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
-import javax.sql.DataSource;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * The general layout of the stairway database tables is:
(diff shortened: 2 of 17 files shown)


src/main/java/bio/terra/stairway/HookWrapper.java
```diff
@@ -1,60 +1,126 @@
...
+          operation.handleStepHook(context, stepHook);
+        } catch (Exception ex) {
+          logger.warn("Step Hook failed with exception", ex);
+        }
+      }
     }
-    return HookAction.CONTINUE;
-  }
-
-  private FlightContext makeCopy(FlightContext fc) {
```
yonghaoy, line 47: Just to confirm, we won't new a copy for FlightContext to handleHook?

src/main/java/bio/terra/stairway/HookWrapper.java
```diff
@@ -1,60 +1,126 @@
...
-    return handleHook(flightContext, context -> stairwayHook.startStep(context));
+  void startFlight(FlightContext flightContext) {
+    handleHookList(flightContext, HookOperation.START_FLIGHT);
   }
 
-  public HookAction endStep(FlightContext flightContext) {
-    return handleHook(flightContext, context -> stairwayHook.endStep(context));
+  void startStep(FlightContext flightContext) throws InterruptedException {
+    // First handle plain step hooks
+    handleHookList(flightContext, HookOperation.START_STEP);
```
yonghaoy, line 75: What is the "best practice" recommandation for using `stepFactory` vs `plain step hooks`. 
Will (1) those two stay together just like this or (2) `stepFactory`  may replace `plain step hooks` some day.
If (2), might be helpful if we mark `plain step hooks`  `startStep` and `endStep` as `@deprecated`?

yonghaoy (approved):
Thanks Dan! This is really helpful!

src/main/java/bio/terra/stairway/FlightContext.java
```diff
@@ -11,8 +11,7 @@
 public class FlightContext {
   private Stairway stairway; // the stairway instance running this flight
   private String flightId; // unique id for the flight
-  private final String
-      flightClassName; // class name of the flight;  for recreating the flight object
+  private final String flightClassName; // class name of the flight
```
pshapiro4broad: nit: these end of line comments should really be field javadocs

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -355,18 +359,22 @@ private void updateFlightState(String comment, String sql, String flightId, Stri
...
         commitTransaction(connection);
+        hookWrapper.stateTransition(flightContext);
         break;
       } catch (SQLException ex) {
         if (!retrySqlException(ex)) {
           throw new DatabaseOperationException(
-              "Failed to update status to " + status + " for flight: " + flightId, ex);
+              String.format(
+                  "Failed to update status to %s for flight: %s",
+                  flightContext.getFlightStatus().name(), flightContext.getFlightId()),
```
pshapiro4broad: `.name()` is redundant as `FlightStatus` is an `enum` so its default String representation is its name.

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -392,20 +405,40 @@ void disownRecovery(String stairwayId) throws DatabaseOperationException, Interr
 
     for (int retry = 0; retry < SERIALIZATION_RETRIES; retry++) {
       try (Connection connection = dataSource.getConnection();
-          NamedParameterPreparedStatement statement =
-              new NamedParameterPreparedStatement(connection, sql);
+          NamedParameterPreparedStatement getStatement =
+              new NamedParameterPreparedStatement(connection, sqlGet);
+          NamedParameterPreparedStatement updateStatement =
```
pshapiro4broad, line 410: Since it's only used in the `if` below you could move this declaration and initialization (and `sqlUpdate`'s) inside that block.

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -629,6 +652,65 @@ FlightContext resume(String stairwayId, String flightId)
...
+    return null;
+  }
+
+  // The results set must be positioned at a row from the flight table,
+  // and the select list must include debug_info, class_name, and status.
+  private FlightContext makeFlightContext(Connection connection, String flightId, ResultSet rs)
+      throws InterruptedException, DatabaseOperationException, SQLException {
+    List<FlightInput> inputList = retrieveInputParameters(connection, flightId);
+    FlightMap inputParameters = new FlightMap(inputList);
+    FlightDebugInfo debugInfo = null;
```
pshapiro4broad, line 696: I think a better idiom here is to not initialize and declare the variable as `final`. That way it must be initialized and you don't need to come up with a default value.
```suggestion
    final FlightDebugInfo debugInfo;
```

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -629,6 +652,65 @@ FlightContext resume(String stairwayId, String flightId)
...
+              new NamedParameterPreparedStatement(connection, sqlGetFlight)) {
+
+        startTransaction(connection);
+
+        getFlightStatement.setString("flightId", flightId);
+        FlightContext flightContext = null;
+        try (ResultSet rs = getFlightStatement.getPreparedStatement().executeQuery()) {
+          if (rs.next()) {
+            flightContext = makeFlightContext(connection, flightId, rs);
+          }
```
pshapiro4broad, line 675: What if `rs.next()` is false? Do you want to throw a specific exception in that case?

src/main/java/bio/terra/stairway/Stairway.java
```diff
@@ -162,7 +163,7 @@ public Builder enableWorkQueue(boolean enableWorkQueue) {
      * @return this
      */
     public Builder keepFlightLog(boolean keepFlightLog) {
-      this.keepFlightLog = Boolean.valueOf(keepFlightLog);
+      this.keepFlightLog = keepFlightLog;
```
pshapiro4broad, line 166: It looks like `Builder.keepFlightLog` could be changed to `boolean` if you initialize it with a default value of `true`, since that is the value it uses if the field is `null` in the `Stairway` constructor.

src/main/java/bio/terra/stairway/Stairway.java
```diff
@@ -181,12 +182,18 @@ public Builder flightFactory(FlightFactory flightFactory) {
...
+     * order in which they are added to the builder.
+     *
      * @param stairwayHook object containing hooks for logging at beginning and end of flight and
      *     step of stairway
      * @return this
      */
     public Builder stairwayHook(StairwayHook stairwayHook) {
-      this.stairwayHook = stairwayHook;
+      if (this.stairwayHooks == null) {
+        this.stairwayHooks = new ArrayList<>();
```
pshapiro4broad: IMO I would initialize the field `Builder.stairwayHooks` as `new ArrayList<>()`. That would let you remove the null check here and in the `HookWrapper` constructor.

src/test/java/bio/terra/stairway/fixtures/TestUtil.java
```diff
@@ -18,17 +18,18 @@
...
 import org.slf4j.LoggerFactory;
 
 public final class TestUtil {
-  private static Logger logger = LoggerFactory.getLogger(TestUtil.class);
+  private static final Logger logger = LoggerFactory.getLogger(TestUtil.class);
 
   private TestUtil() {}
 
-  public static final Integer intValue = Integer.valueOf(22);
+  public static final Integer intValue = 22;
```
pshapiro4broad: The types don't need to be boxed here
```suggestion
  public static final int intValue = 22;
```

src/test/java/bio/terra/stairway/fixtures/TestUtil.java
```diff
@@ -18,17 +18,18 @@
...
   public static final String strValue = "testing 1 2 3";
-  public static final Double dubValue = new Double(Math.PI);
+  public static final Double dubValue = Math.PI;
```
pshapiro4broad:
```suggestion
  public static final double dubValue = Math.PI;
```

src/test/java/bio/terra/stairway/fixtures/TestUtil.java
```diff
@@ -82,19 +93,32 @@ public static void sleepPause() throws InterruptedException {
...
             .stairwayName(stairwayName)
             .workQueueProjectId(projectId)
             .enableWorkQueue(enableWorkQueue)
-            .maxParallelFlights(2)
-            .build();
+            .maxParallelFlights(2);
+
+    for (int i = 0; i < hooks; i++) {
+      int hookId = i + 1;
+      TestHook hook = new TestHook(Integer.toString(hookId));
```
pshapiro4broad: Using `String.valueOf(x)` everywhere you use `XType.toString(x)` is generally better IMO because if the type of `x` ever changes you don't need to change this part of the code.

src/test/java/bio/terra/stairway/fixtures/TestUtil.java
```diff
@@ -106,17 +130,27 @@ private static Stairway makeStairway(
...
+    Stairway.Builder builder =
         Stairway.newBuilder()
             .stairwayClusterName("stairway-cluster")
             .stairwayName(stairwayName)
             .workQueueProjectId(null)
-            .maxParallelFlights(2)
-            .build();
+            .maxParallelFlights(2);
+
+    for (int i = 0; i < hooks; i++) {
```
pshapiro4broad, line 147: Not sure if this is better or worse but you can also write this using java streams, if you wanted to pass a list to the builder directly.
```
        List<TestHook> hooks = 
                IntStream.range(0, numHooks)
                .map(x -> x + 1)
                .mapToObj(String::valueOf)
                .map(TestHook::new)
                .collect(Collectors.toList());
```

pshapiro4broad (commented):
Just a bunch of minor things, feel free to take/ignore as you see fit. Although you may want to explicitly handle the case where an expected flight can't be found.

src/main/java/bio/terra/stairway/HookWrapper.java
ddietterich: Yes. The old code was not making a very good copy anyway. It was only a surface copy.
In thinking about it, I realized that I trust the Flight and Step code with the live context, so I might as well trust the Hook code as well. I added [PF-373](https://broadworkbench.atlassian.net/browse/PF-373) to track the issue.

src/main/java/bio/terra/stairway/HookWrapper.java
ddietterich: I wasn't planning to deprecate. If you need to hold state, then use the factory. Otherwise, do it the simple way.

src/main/java/bio/terra/stairway/FlightDao.java
ddietterich: To maintain the existing Flight constructor, we used a setter to populate the debugInfo when it is being used.
So I think it has to be not `final`

src/main/java/bio/terra/stairway/FlightDao.java
ddietterich: No. We commit the transaction and return null.

src/main/java/bio/terra/stairway/Stairway.java
ddietterich: Yes. I'm doing premature optimization by avoiding a tidbit of code in the case where no hooks are defined. Since it looks like we will always have at least trace hooks defined in production, that is entirely useless.

src/main/java/bio/terra/stairway/FlightDao.java
pshapiro4broad: In that case a NPE will be triggered, since the next line after the call to this method passes the `null` result to `queueFlight()` which calls `flightContext.getFlightId()`.

src/main/java/bio/terra/stairway/FlightDao.java
pshapiro4broad: Sorry I wasn't clear; I mean that the local variable (not the `FlightContext` field) can be declared final. Once you do that, you no longer need the `= null` initializer. Nothing else in this method would change.

PR #54, opened 2021-01-19
Use the BroadBot service account to perfom version bumps on master

choover-broad:
Pushing to protected branches as the GitHub actions user is not supported: https://github.community/t/how-to-push-to-protected-branches-in-a-github-action/16101/29

So, let's try using our BroadBot service account instead. I've added the account's PAT as a repo secret and updated repo settings to allow it to push to the develop branch.


ddietterich (approved):
Okee dokee

PR #56, opened 2021-02-01
PF-406: Add a FlightHook with factory like the StepHook

ddietterich:
Parallel implementation to the StepHook. I considered making an incompatible change and deriving both StepHook and FlightHook from a common class, but decided it wasn't enough code to be concerned with.


src/main/java/bio/terra/stairway/FlightHook.java
```diff
@@ -0,0 +1,7 @@
+package bio.terra.stairway;
+
+public interface FlightHook {
```
wchamber: Worth a class comment on what this is used for and when start/endFlight get called? I'm always a little surprised that they can be called multiple times on a single flight id.

src/main/java/bio/terra/stairway/StairwayHook.java
```diff
@@ -26,4 +26,8 @@ default HookAction stateTransition(FlightContext context) throws InterruptedExce
   default Optional<StepHook> stepFactory(FlightContext context) throws InterruptedException {
     return Optional.empty();
   }
+
+  default Optional<FlightHook> flightFactory(FlightContext context) throws InterruptedException {
```
wchamber: optional: could use some more comments on the StairwayHook interface about what it's used for, and any guarantees about when its functions will be run.

wchamber (approved):
thanks!

pshapiro4broad:
> I considered making an incompatible change and deriving both StepHook and FlightHook from a common class, but decided it wasn't enough code to be concerned with.

~How many lines of code (outside stairway itself) would be affected by this incompatible change? The existing hook mechanism has only been out for two weeks. I would recommend making the breaking change and refactor the code. Not because of the volume of code, but for the mental overhead of looking at apparently identical code and needing to know that they are actually doing the same thing, but have different code paths due to backward compatibility.~ Never mind, I see that the shared code is in `StairwayHook`, not `StepHook`.

If you're firm on not changing the StepHook API, maybe there's a way to introduce a new base interface class that's compatible with the existing API.

ddietterich:
@wchamber What do you think about refactoring so there is only one kind of factory hook instead of two, with methods `start` and `end`? I think you have the only code that is using the interface right now.

pshapiro4broad:
Not sure, but it looks like you may be able to get away with making `StepHook` and `FlightHook` extend `StairwayHook`. I think that would let you simplify `HookOperation` and maybe refactor the various `handleHookList` methods.

wchamber:
I don't mind changing my existing code at all, so that's no cost from my point of view.
I'd find it clearer if step hooks & flight hooks were different classes but they're fine to share an interface. Not a strong preference.
I do need different instances for each flight & each step. It might be a good idea to have `FlightHooks` create `StepHooks` but I don't need that at the moment with implicit context propagation for traces. (i.e. A FlightHook might have some shared state that every StepHook it creates and only those StepHooks can refer to.)

I'd find it confusing if the StepHook and FlightHook extend StairwayHook as they would have inherited methods that would never be called by the HookWrapper.

ddietterich:
Just to be clear, the idea is to have the StairwayHook object have two factory methods like they do:
```
  default Optional<DynamicHook> stepFactory(FlightContext context) throws InterruptedException {
    return Optional.empty();
  }

  default Optional<DynamicHook> flightFactory(FlightContext context) throws InterruptedException {
    return Optional.empty();
  }
```
And have one DynamicHook class like:
```
public interface DynamicHook {
  HookAction start(FlightContext context) throws InterruptedException;

  HookAction end(FlightContext context) throws InterruptedException;
}
```
That would save one class and some duplicate code in the HookWrapper.

wchamber:
sgtm

src/main/java/bio/terra/stairway/Flight.java
```diff
@@ -106,14 +107,14 @@ public void run() {
...
       flightExit(flightStatus);
-      hookWrapper().endFlight(flightContext);
     } catch (InterruptedException ex) {
       // Shutdown - try disowning the flight
       logger.warn("Flight interrupted: " + context().getFlightId());
       flightExit(READY);
     } catch (Exception ex) {
       logger.error("Flight failed with exception", ex);
     }
+    hookWrapper().endFlight(flightContext);
```
wchamber: unlike `hookWrapper().startFlight`, this isn't wrapped in any `try catch`. Just want to verify that the asymmetry is intended.

src/main/java/bio/terra/stairway/HookWrapper.java
```diff
@@ -66,34 +79,48 @@ void checkHookAction(HookAction action) {
...
+    handleDynamicHookList(flightContext.getStepHooks(), flightContext, HookOperation.END_STEP);
     flightContext.setStepHooks(null);
   }
 
   void endFlight(FlightContext flightContext) {
+    // First handle plain flight hooks
     handleHookList(flightContext, HookOperation.END_FLIGHT);
+    // Next handle the flight hooks list from the flight context
+    handleDynamicHookList(flightContext.getFlightHooks(), flightContext, HookOperation.END_FLIGHT);
+    flightContext.setFlightHooks(null);
```
wchamber, line 123: it's a little unusual to me that the flight hooks are removed from the FlightContext, but I don't have much of a reason for keeping them, just a possible bug if they get examined after `endFlight`. I wouldn't particularly expect `HookWrapper#endFlight` to modify the FlightContext this way, though it does match `startFlight`. Any particular reason for removing them?

src/main/java/bio/terra/stairway/FlightHook.java
wchamber: I think FlightHook can be removed now?

wchamber (approved):
couple tiny questions. Thanks for the extensive documentation on StairwayHook!
Also, update the PR description for the recent changes?

yonghaoy (approved):

src/main/java/bio/terra/stairway/StairwayHook.java
ddietterich: I added documentation to StairwayHook class and tried to provide cardinality info.

src/main/java/bio/terra/stairway/Flight.java
ddietterich: Good point, I wrapped it. There is nothing to be done besides log on an exception, but probably good policy not to send exceptions out to the thread pool.

src/main/java/bio/terra/stairway/HookWrapper.java
ddietterich: No strong reason. It was symmetric with endStep logic. I wouldn't mind an NPE if some code assumed that list was always there when it was no longer valid.

PR #57, opened 2021-02-07
[PF-369] Add retry to all database requests

ddietterich:
There are two main parts of this PR. First, I created StairwayInstanceDao to handle interactions with the StairwayInstance table. The FlightDao was getting large and those methods separate fairly cleanly. That required factoring out the common transaction methods into DbUtils class.

Second, I made a generic retry class (DbRetry) and applied it across both DAOs.

This should allow TDR to ride through transient SqlProxy failures and serialization contention.

The Stairway unit tests pass. I also ran the Data Repo connected tests to further sanity check.


src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -55,144 +56,43 @@
...
-    } catch (SQLException ex) {
-      throw new DatabaseOperationException("Stairway instance enumeration failed", ex);
-    }
-    return instanceList;
-  }
-
-  /** Record a new flight */
   void submit(FlightContext flightContext)
       throws DatabaseOperationException, InterruptedException, DuplicateFlightIdSubmittedException {
+    DbRetry dbRetry = new DbRetry("flight.submit");
```
pshapiro4broad: For the pattern `new Foo().method()` (that is, the case where you create an object for the sole purpose of calling a single method on it) I think it's clearer to use a static method on the object. You can still create an object in that static method, but the caller doesn't need to know about this detail.

src/main/java/bio/terra/stairway/DbRetry.java
```diff
@@ -0,0 +1,69 @@
...
+      logger.info("Caught SQL connection or resource error (" + ss + ") - retrying");
+      return true;
+    }
+    return false;
+  }
+
+  private void retryWait() throws InterruptedException {
+    int sleepMS = ThreadLocalRandom.current().nextInt(WAIT_MIN_MS, WAIT_MAX_MS);
+    TimeUnit.MILLISECONDS.sleep(sleepMS);
+    logger.info("retrying for {}", logString);
```
pshapiro4broad: IMO this seems like `debug` not `info`. Unless you believe retrying will be relatively uncommon.

src/main/java/bio/terra/stairway/DbFunction.java
```diff
@@ -0,0 +1,14 @@
...
+import bio.terra.stairway.exception.DatabaseOperationException;
+import java.sql.SQLException;
+
+/**
+ * We provide a custom functional interface to set the right thrown exceptions
+ *
+ * @param <R> The return value of the database function. If there is no return value, specify Void.
+ */
+@FunctionalInterface
+public interface DbFunction<R> {
```
pshapiro4broad: Since this is only used in the class `DbRetry` I don't think there's a benefit to declaring in a top level scope. Declaring it local to `DbRetry` would have the same effect and not pollute the global namespace.

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -55,144 +56,43 @@
...
+    dbRetry.perform(() -> submitInner(flightContext));
+  }
+
+  private Void submitInner(FlightContext flightContext)
```
pshapiro4broad: I'm not sure how well this would work, but you could have two `perform` methods, with one that supports `void` methods. That would avoid the need for `Void` and `return null;`

src/main/java/bio/terra/stairway/DbRetry.java
```diff
@@ -0,0 +1,69 @@
...
+  private static final Logger logger = LoggerFactory.getLogger(DbRetry.class);
+  private static final int MAX_RETRIES = 20;
+  private static final int WAIT_MIN_MS = 250;
+  private static final int WAIT_MAX_MS = 1000;
+  private static final String PSQL_SERIALIZATION_FAILURE = "40001";
+  private static final String PSQL_DEADLOCK_DETECTED = "40P01";
+  private static final String PSQL_CONNECTION_ISSUE_PREFIX = "08";
+  private static final String PSQL_RESOURCE_ISSUE_PREFIX = "53";
+
+  private final String logString;
```
wchamber, line 44: add a quick comment on what this is or should be used for

src/main/java/bio/terra/stairway/DbUtils.java
```diff
@@ -0,0 +1,23 @@
+package bio.terra.stairway;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+
+/** Database methods shared across DAOs */
+public class DbUtils {
+  static void startTransaction(Connection connection) throws SQLException {
```
wchamber: Does this class need to be public if it's functions are package private? ditto on DbRetry

src/main/java/bio/terra/stairway/StairwayInstanceDao.java
```diff
@@ -0,0 +1,191 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Database operations on the Stairway instance table */
+@SuppressFBWarnings(
+    value = "RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE",
+    justification = "Spotbugs doesn't understand resource try construct")
+public class StairwayInstanceDao {
+  private static final Logger logger = LoggerFactory.getLogger(StairwayInstanceDao.class);
+  static final String STAIRWAY_INSTANCE_TABLE = "stairwayinstance";
```
wchamber: can this be private?

src/main/java/bio/terra/stairway/StairwayInstanceDao.java
```diff
@@ -0,0 +1,191 @@
...
+import java.util.List;
+import javax.sql.DataSource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Database operations on the Stairway instance table */
+@SuppressFBWarnings(
+    value = "RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE",
+    justification = "Spotbugs doesn't understand resource try construct")
+public class StairwayInstanceDao {
```
wchamber: ditto does this class need to be public or just package private?

wchamber (approved):

src/main/java/bio/terra/stairway/FlightDao.java
ddietterich: From what I read, an `@FunctionalInterface` has to have a return value. The recommendation was to return Void, but you cannot `new` a Void, so you return null.

Is there a variant of functional interface that does not return a value?

src/main/java/bio/terra/stairway/DbRetry.java
```diff
@@ -0,0 +1,69 @@
...
+ * backoff with a maximum number of retries. At this time, there is no use case for setting
+ * different retry times for different calls. If there is, then an additional constructor can be
+ * added to allow specification of what are coded as statics.
+ */
+public class DbRetry {
+  private static final Logger logger = LoggerFactory.getLogger(DbRetry.class);
+  private static final int MAX_RETRIES = 20;
+  private static final int WAIT_MIN_MS = 250;
+  private static final int WAIT_MAX_MS = 1000;
+  private static final String PSQL_SERIALIZATION_FAILURE = "40001";
```
yonghaoy, line 38: One of the exception I saw from log is 
`org.postgresql.util.PSQLException: An I/O error occurred while sending to the backend.` [[log]](org.postgresql.util.PSQLException: An I/O error occurred while sending to the backend.)
I didn't find the number for that exception. Does it covered by any of those code?

yonghaoy (approved):

src/main/java/bio/terra/stairway/FlightDao.java
pshapiro4broad: Not sure where this info is in the java spec, but `void` must be allowed as `Consumer` is a functional interface.

src/main/java/bio/terra/stairway/StairwayInstanceDao.java
ddietterich: I made a pass through all of the classes to clean up public classes that shouldn't be. Thanks for picking up on that.

src/main/java/bio/terra/stairway/DbRetry.java
ddietterich: I believe that will be a SQLstate beginning with PSQL_CONNECTION_ISSUE_PREFIX ("08").
The SQLstate is a member of the PSQLException object, so if you have logging that prints the whole object, you should have that value.

src/main/java/bio/terra/stairway/FlightDao.java
pshapiro4broad: FYI here's the relevant part of the JLS https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.8

ddietterich: Got it. It is a little bit of repetition in the DbRetry class, but gets rid of the ugly `return null` in the main code. Seems like a good trade-off.

PR #61, opened 2021-03-11
PF-564 Restore public accessors in QueueMessage files

ddietterich:
Serializing doesn't work if Jackson cannot see the data...

I added a unit test for the case I also found an ordering dependency in the connected tests. Fixed by separating the tests into separate modules.


yonghaoy (approved):
Thanks!

yonghaoy:
And I think Tova is out this week

yonghaoy:
Also, one thing that might be help for debuging, but may not safe is printing the message that are failed to serialize.
https://github.com/DataBiosphere/stairway/blob/develop/src/main/java/bio/terra/stairway/QueueMessage.java#L34

src/main/java/bio/terra/stairway/QueueMessageNoFields.java
```diff
@@ -4,14 +4,14 @@
 
 // We use this class to read the type information, to guide deserializing the right type
 @JsonIgnoreProperties(ignoreUnknown = true)
-class QueueMessageNoFields extends QueueMessage {
+public class QueueMessageNoFields extends QueueMessage {
```
pshapiro4broad: I'm pretty sure the class doesn't need to be public, just the fields.

yonghaoy (approved):

jmczerk (approved):

PR #62, opened 2021-03-19
[PF-580] Make Stairway Treat Flight Parameters as Opaque Strings

jmczerk:
Still some docs and testing to be done on this, but in case anyone wants an early look...

The idea here is to decouple the serialization/deserialization of Flight parameters from Stairway's own internals, by allowing the consumer to pass ser/des objects to `get`/`put` methods of `FlightMap`.  Internally the database representation of the `FlightMap` becomes a mapping of `String` type keys to opaque `String` objects.

In order to maintain backward compatibility, the current `get`/`put` methods are maintained, which respectively call new public overloads of each which take objects implementing ser/des interfaces.  The existing methods in turn construct and pass default impls of the ser/des interfaces which maintain the implicit Jackson serialization behavior under the hood, and call the new methods with them.

In order to support upgrade, a new column is added to the `flight` and `flightlog` tables with a parameter version; this defaults to version 1 at upgrade and will be written as version 2 for subsequent writes.  This allows for backward compatibility of these tables by passing a version to the `fromJson` method of `FlightMap`.  

Note that while backward compatibility is maintained for Flights proper when on-disk format of parameters change (assuming the parameter does the right things), queries on `flightinput` will not be backward compatible.


build.gradle
```diff
@@ -75,6 +75,9 @@ dependencies {
     // For log control during unit testsing
     testImplementation group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.3'
     testRuntime group: 'org.codehaus.groovy', name: 'groovy', version: '2.5.7'
+
+    // Enable AutoValue in test code
+    testAnnotationProcessor group: "com.google.auto.value", name: "auto-value", version: "1.7.4"
```
ddietterich: I do not want to use AutoValue in Stairway. If you want to, you can use Immutables. I would prefer neither, as I find it makes the code opaque.

local-dev/run_postgres.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
```
ddietterich: Do we really need a cut-n-paste of this script in every repo?
Seems like a maintenance headache.

src/main/resources/stairway/db/changesets/20210318_metadata_version.yaml
```diff
@@ -0,0 +1,21 @@
+databaseChangeLog:
+  - changeSet:
+      id: metadata_version
+      author: jc
```
ddietterich: With you and Jay, this is no longer unique. Doesn't matter, just an observation.

src/main/java/bio/terra/stairway/DefaultFlightParameterDeserializer.java
```diff
@@ -0,0 +1,24 @@
+package bio.terra.stairway;
+
+import bio.terra.stairway.exception.JsonConversionException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.io.IOException;
+
+public class DefaultFlightParameterDeserializer<T> extends FlightParameterDeserializer<T> {
```
ddietterich: So I have to construct one of these for each type I want to deserialize? That seems awkward. I wonder if there is a way to do it using interfaces instead of subclassing.

How are the common Java types handled? String, Int, Long are the most frequent parameters.

src/main/java/bio/terra/stairway/FlightFilter.java
```diff
@@ -105,8 +105,16 @@ public FlightFilter addFilterFlightStatus(FlightFilterOp op, FlightStatus status
...
    * @throws FlightFilterException if key is not supplied
    */
+  @Deprecated
   public FlightFilter addFilterInputParameter(String key, FlightFilterOp op, Object value)
       throws FlightFilterException {
+    return addFilterInputParameter(
+        key, op, value, new DefaultFlightParameterSerializer(StairwayMapper.getObjectMapper()));
+  }
+
+  public FlightFilter addFilterInputParameter(
```
ddietterich: Maybe a better API for this would be to take `String serializedObject` instead of a serializer.
It doesn't look like you are using the serializer, so I don't think this will work as is.

It does put a requirement on the serdes to generate exactly the same string format for the same object. I wonder if that is a problem.

I guess the alternative would be to always retrieve all of the data from the database and filter it in Java code instead o fin the database. That would allow us to have a comparison operation - or deserialize from the stored form and use the `object.equals`. That would give the opportunity for the serdes to fix version mismatches.

src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -5,24 +5,86 @@
...
 /**
  * FlightMap wraps a {@code HashMap<String, Object>} It provides a subset of the HashMap methods. It
  * localizes code that casts from Object to the target type. It provides a way to set the map to be
  * immutable.
  */
 public class FlightMap {
-  private Map<String, Object> map;
+  @VisibleForTesting
+  static final class ObjectContainer {
+    private final String data;
```
ddietterich: This is the serialized form, right? A more descriptive name would have helped me understand the code more easily.

src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -5,24 +5,86 @@
...
+      this.serializer = serializer;
+    }
+
+    public String getData() {
+      // Constructors ensure that one can only have valid data OR object+serializer.
+      return (data != null) ? data : serializer.serialize(object);
+    }
+
+    public <T> T getObject(FlightParameterDeserializer<T> deserializer) {
+      return (object != null) ? deserializer.safeCast(object) : deserializer.deserialize(data);
```
ddietterich: Is immutability worth repeated deserialization?
I think it'd be better to store the resulting object.

Similarly, at each step boundary, we will serialize the working map, but won't save the serialization work in `data`.

src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -5,24 +5,86 @@
...
+    public ObjectContainer convert(String value) {
+      return new ObjectContainer(value);
+    }
+  }
+
+  private Map<String, ObjectContainer> map;
 
   public FlightMap() {
     map = new HashMap<>();
+
```
ddietterich: A bit of documentation about what you are doing here would be helpful.

ddietterich (commented):
There are a few issues in here. My top two are the structure of serializers and the application of filters in enumeration.

src/main/java/bio/terra/stairway/FlightMap.java
jmczerk: My first pass cached a reference to the object on deserialization in the call to `getObject()`, but that raised some ambiguity around what the right behavior would be at serialization time if the referenced object and the serialized data have diverged from one another.

It sounds like I took the exact wrong approach though; would the preferable approach be that the referenced object is the object-of-record and the serialized data member is only valid until the object gets populated?  If so, renaming `data` to `initialSerializedState` would make this clear when I change the behavior.

The implication would be that any changes to the referenced object between `put` or `get` and serialization will be what is recorded at serialization time.

ddietterich: Yet another alternative (with some downsides) is to only keep the serialized form. When the object is put, we immediately serialize it. When it is gotten, the deserializer is provided and we make the object. We never try to keep track of the serializer in the mao. That would reduce the map to `<String, String>`

One benefit of that approach is the normal (no fail overs path) would break if the serdes was broken. These days we only find out about it when there is a failure and the deserialization fails. Not so great.

The bad part is we would spend more time doing deserialization. We already serialize everything at each step boundary.

src/main/java/bio/terra/stairway/DefaultFlightParameterDeserializer.java
jmczerk: No, the existing `get()` API instantiates one of these under the hood and calls the new overload with it.  Nobody except `FlightMap` (and maybe `...Filter...`, depending on how that resolves) should be instantiating this directly.  Classes that want custom ser/des will need to implement to `FlightParameterSerializer/FlightParameterDeserializer<T>`.

As to why `FlightParameterDeserializer<T>` is an abstract and not an interface, I'll add a comment on that class in the review, as there is a specific trade-off that this enables.

Common types can just continue to use the old API.  I'll drop a comment into the review for `FlightMapTest.java` with the resulting JSON for a mix of custom and default serialized classes/primitives.

src/test/java/bio/terra/stairway/FlightMapTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+    FlightsTestNonPojo nonPojoIn = new FlightsTestNonPojo(1.25f);
+    outMap.put("mynonpojo", nonPojoIn, FlightsTestNonPojo.serializer());
+
+    Integer intIn = 3;
+    outMap.put("myint", intIn);
+
+    String json = outMap.toJson();
+    logger.info("JSON: '{}'", json);
+
```
jmczerk, line 32: Prettified version of the JSON created by the call to `outMap.toJson()`:

```
[
    "java.util.HashMap",
    {
        "myint": "3",
        "mynonpojo": "{\"version\":2,\"uuid\":\"a5f68d4d-eaa7-4055-bbe1-882428f4ab4a\",\"value\":1.25}",
        "mypojo": "[\"bio.terra.stairway.fixtures.FlightsTestPojo\",{\"astring\":\"mystring\",\"anint\":99}]"
    }
]

```
Note that the `put()` calls for `pojoIn` and `intIn` use default API, where call for `nonPojoIn` calls new custom serializer overload.

src/main/java/bio/terra/stairway/FlightMap.java
jmczerk: From a code complexity standpoint, this would be a big win, as it would eliminate a bunch of type erasure related messiness that arises from storing references as `Object`, and make the ser/des of the map simpler.  It also makes it perfectly clear that `get` and `put` are the only ways to effect changes to what gets serialized, and eliminates any implicit state changes due to reference mutability.

Does "serializing everything at each step boundary" imply that `toJson()` would be expected to be called more frequently per-parameter than `get()/put()`?  If so, it sounds like this might be an efficiency win too, as we'd only serialize once, and then reuse the serialized data.  We should also figure in that we'll be paying for a deserialization on the first `get()` anyway...

ddietterich: In the current scheme, the input parameters are serialized at flight launch and are only deserialized again if the flight is queued, either directly or if there is a failover and the flight needs to be restarted. The working map is serialized at the end of each step is deserialized on flight restart.

Just keeping the serialized state in the map is probably a wash. There would be more deserialization and less serialization. Typically, working map entries are written once and then read one or more times.

jmczerk:
Post discussion with @ddietterich, we decided to implement the approach where `FlightMap` only stores serialized strings.  This significantly reduces code complexity, at the cost of doing more deserializations.  It also removes ambiguity around whether changes to mutable objects participating in `get()/put()` will be implicitly tracked, since we no longer hold references to mutable objects.

Main areas of changes are:

### FlightMap
* New `get()` and `put()` overloads are added to `FlightMap` which add a `FlightParameterDeserializer<T>` and `FlightParameterSerializer<T>` argument, respectively.
  * At `put` time, the passed object is serialized to a `String` using the passed serializer and stored in a `Map<String, String>`.
  * At `get` time, the `String` stored at the passed key is deserialized to an object using the passed deserializer.
* Existing `get()` and `put()` methods construct a `DefaultFlightParameterDeserialzer<T>` and `DefaultFlightParameterSerialzer<T>` under the hood, which provide the existing Jackson-based ser/des behavior, and pass them to the new method.
  * Primitive types should continue to use the existing `get()/put()` methods going forward.  
  * We might consider encouraging non-primitive parameter types to implement explicit serializers/deserializers.

### Enumeration
* Existing `FlightFilter#addFilterInputParameter()` behavior is preserved (uses a `DefaultFlightParameterSerializer<T>` under the hood, similar to `FlightMap`).
* Overload of `FlightFilter#addFilterInputParameter()` is added that passes a custom serializer via an additional `FlightParameterSerializer<T>` parameter.
* New method `FlightFilter#addFilterInputParameterRaw()` is added which allows passing raw JSON to queries, which may be useful for debugging when parameter on-disk representation changes.

### DAO
* Tables `flight` and `flightlog` add new “parameters version” columns which are used to differentiate between `FlightMap` on-disk format from previous code, required for `FlightMap` deserialization.  These default to 1 for upgrades, and current code version will write new flights with version 2.

src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -67,30 +59,61 @@ public void makeImmutable() {
...
+   * that type.
    *
    * @param <T> - type of class to expect in the hash map
    * @param key - key to lookup in the hash map
    * @param type - class requested
    * @return null if not found
    * @throws ClassCastException if found, not castable to the requested type
    */
+  @Nullable
   public <T> T get(String key, Class<T> type) {
```
wchamber, line 73: Should we `@Deprecated` the existing Object serialization? If we think the current version makes it difficult to introduce changes, and we're adding a new pair of methods, should we encourage clients to migrate to them by marking the existing functions as deprecated? If we don't, what would we want from the new functions to encourage users to use them?
We could add explicit overloads for the primitives we want to support, e.g. `String get(String key)`, `void put(String key, String value)`.

wchamber (commented):
My main question looking at this is do we want to add the serialization to the FlightMap API, or do we just want serialization to be done externally and leave FlightMap unaware that the resulting string is the serialized form of something?
We don't need to add the `FlightParameter[De]serializer` interfaces, we can do our own serialization before calling `flightMap.<String>put(myKey, mySerializedValue)`. That would save us from needing to change the existing database, or the rest of the PR I think.
If Stairway serialization is really it's own layer, it's not unreasonable for `MyClass` to have a `putInFlightMap` and `getFromFlightMap` method that abstracts how the serialization is done, what the FlightMap keys are, how to do schema evolution. The explicit FlightParameterSerializer interface then seems less helpful.
I think there's still an argument for deprecating the `Object` get/put, and promoting using the String primitive.

OTOH, making serialization more explicit in the FlightMap API might be useful in and of itself.
@ddietterich thoughts?

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -94,8 +95,10 @@ private void submitInner(FlightContext flightContext)
     final String sqlInsertFlight =
         "INSERT INTO "
             + FLIGHT_TABLE
-            + " (flightId, submit_time, class_name, status, stairway_id, debug_info)"
-            + "VALUES (:flightId, CURRENT_TIMESTAMP, :className, :status, :stairwayId, :debugInfo)";
+            + " (flightId, submit_time, class_name, status, stairway_id, debug_info, "
+            + "    output_parameters_version)"
+            + "VALUES (:flightId, CURRENT_TIMESTAMP, :className, :status, :stairwayId, "
+            + " :debugInfo, :outputParametersVersion)";
```
yonghaoy, line 101: this remindes me a breakage in RBS last time when after Stairway adding `debuginfo` here. 
What happens last time is the older flight(in RUNNING or ERROR state) dose not have debugInfo field in db. But after we upgrade Stairway version and restart the server, Stairway tries to recover those old flights, and it is looking for `debuginfo` from db(I forgot which table, might be `flightlog` table). Then it gets NPE for some of following operations when using them.  It happens a while ago and I forgot the details, just want to make sure we can handle this case gracefully

jmczerk: Good point @yonghaoy ...  Both of the new fields are added with default values of 1, so that at database upgrade time existing rows will have this added with version 1.  The version in code is set to 2, so newly written rows will use 2.

I have also created a process for testing database upgrades by storing static data produced with earlier versions of code as static test resources under source control, but I didn't want to add too much to this PR.  That will be a follow-on once this merges.

The tests do use utility methods to make sure upgrade happens at Stairway init time; it would be good to sanity check that "real" Stairway instances will upgrade the database as well.

src/main/java/bio/terra/stairway/FlightMap.java
jmczerk: I'm on board with this, though the `get` overloads would only differ by return type, so would need to become `getInt()`, `getString()`... 

That said, if we do that I'd prefer to make that a follow-on effort so that we can move the database changes along.

jmczerk:
Regarding @wchamber suggestion of just putting/getting `String` instances, I’m not sure I agree, but I’m having a hard time coming up with a strong factual argument.  

I feel like just using `String` is more likely to create disparate ad-hoc serdes logic, whereas the `FlightParameterSerializer`/`FlightParameterDeserializer` interfaces provide some degree of guardrails (if even just by making people think about things a little more intentionally), and are not onerous to implement against. 

I think this is another decision we could revisit post merge, as I'm most concerned with propagating the database changes (which will support either convention).

src/main/java/bio/terra/stairway/FlightMap.java
ddietterich: I do not think we should deprecate this now, until we have agreed to what we want people to migrate to. I think there is a lot of value in having a default serdes and there are other ways to handle not-upward-compatible changes than writing your own serdes.

One thing I had not considered before, but this interface is returning String. To be open ended on what kind of serdes we support, should we change this to binary? Or is everyone comfortable with a character-based serdes? That would be a more disruptive change to the Stairway database.

src/main/java/bio/terra/stairway/DefaultFlightParameterDeserializer.java
```diff
@@ -0,0 +1,41 @@
...
+  /**
+   * Constructs a DefaultFlightParameterDeserializer<T> instance, which uses the passed Jackson
+   * ObjectMapper to deserialize objects of type T to Strings.
+   *
+   * @param type - type Class<T>, required to pass to Jackson {@link ObjectMapper} at
+   *     deserialization time
+   * @param mapper - Jackson {@link ObjectMapper} used to serialize objects of type T in method
+   *     {@link #deserialize(String)}
+   */
+  public DefaultFlightParameterDeserializer(Class<T> type, ObjectMapper mapper) {
```
wchamber, line 28: Is there ever an ObjectMapper to use other than the Stairway.objectMapper()? Fine to leave the constructor for `@VisibleForTesting` if needed, but consider a factory function for the normal use case.
ditto default serializer.

src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -101,14 +125,30 @@ public String toJson() {
     }
   }
 
-  public void fromJson(String json) {
+  private void fromJsonV1(String json) {
     try {
-      map = getObjectMapper().readValue(json, new TypeReference<Map<String, Object>>() {});
```
wchamber, line 106: One thing to check is if this change can be rolled back, e.g. we find some breaking bug, but some v2 flight maps have already been serialized to the database. Can the old binary handle the v2 flights without totally breaking?
If not, we may want to first roll out a change that can read both v1 & v2, but only writes v1. That first change should be easy to roll back, and we get a little more safety about rolling back the change to write to v2.
OTOH, maybe that's overly paranoid. Though with multi-instance and queuing, it's feasible to have new -> old binary flight map messaging during roll out.

src/test/java/bio/terra/stairway/FlightMapTest.java
```diff
@@ -0,0 +1,81 @@
...
+    outMap.put("mypojo", pojoIn);
+
+    FlightsTestNonPojo nonPojoIn = new FlightsTestNonPojo(1.25f);
+    outMap.put("mynonpojo", nonPojoIn, FlightsTestNonPojo.serializer());
+
+    Integer intIn = 3;
+    outMap.put("myint", intIn);
+
+    String json = outMap.toJson();
+    logger.info("JSON: '{}'", json);
```
wchamber, line 31: consider removing the logger.info's in tests. If there's particularly helpful context for debugging an assertion, add it as a message to the `assertEquals`.

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -56,6 +56,7 @@
   static final String FLIGHT_TABLE = "flight";
   static final String FLIGHT_LOG_TABLE = "flightlog";
   static final String FLIGHT_INPUT_TABLE = "flightinput";
+  static final int FLIGHT_PARAMETERS_VERSION = 2;
```
wchamber, line 59: nit: maybe `CURRENT_PARAMETERS_VERSION`?

src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -101,14 +125,30 @@ public String toJson() {
...
+          getObjectMapper().readValue(json, new TypeReference<Map<String, JsonNode>>() {});
+      for (Map.Entry<String, JsonNode> entry : legacyMap.entrySet()) {
+        map.put(entry.getKey(), entry.getValue().toString());
+      }
     } catch (IOException ex) {
       throw new JsonConversionException("Failed to convert json string to map", ex);
     }
   }
 
+  public void fromJson(String json, int version) {
```
wchamber, line 140: putting the version int in the public API feels unfortunate. Can we keep the `fromJson` without an int version, and make the `fromJsonV1` public but deprecated?

src/main/java/bio/terra/stairway/FlightDao.java
jmczerk: I did a little research, and it looks like all downstream consumers are explicitly configured to do an upgrade migrate at Stairway init time.  

TDR does it by setting `db.migrate.updateAllOnStart=true` in `application.properties`, while the others all set `migrate-upgrade` to `true` in a Stairway config block of `application.yml`:

TDR: https://github.com/DataBiosphere/jade-data-repo/blob/develop/src/main/resources/application.properties#L12

WSM: https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/src/main/resources/application.yml#L83

Buffer: https://github.com/DataBiosphere/terra-resource-buffer/blob/master/src/main/resources/application.yml#L29

Janitor: https://github.com/DataBiosphere/terra-resource-janitor/blob/master/src/main/resources/application.yml#L30

src/test/java/bio/terra/stairway/fixtures/FlightsTestNonPojo.java
```diff
@@ -0,0 +1,118 @@
...
+        JsonGenerator generator = factory.createGenerator(outputStream);
+
+        generator.writeStartObject();
+        generator.writeNumberField("version", 2);
+        generator.writeStringField("uuid", nonPojo.getUuid().toString());
+        generator.writeNumberField("value", nonPojo.getValue());
+        generator.writeEndObject();
+        generator.close();
+        outputStream.close();
+        return outputStream.toString();
```
wchamber, line 87: is the canonical way to serialize an object now? Lots of boilerplate.
If not, what's the canonical recommended way?

src/test/java/bio/terra/stairway/fixtures/FlightsTestNonPojo.java
```diff
@@ -0,0 +1,118 @@
...
+        JsonNode map = mapper.readTree(string);
+
+        int version = map.has("version") ? map.get("version").asInt() : 1;
+        UUID uuid = UUID.fromString(map.get("uuid").textValue());
+
+        float value;
+        if (version < 2) {
+          int numerator = map.get("value_numerator").asInt();
+          int denominator = map.get("value_denominator").asInt();
+          value = (float) numerator / (float) denominator;
```
wchamber, line 106: Worth some comments about this since I don't see any numerators or denominators in the serialize function.

src/test/java/bio/terra/stairway/fixtures/FlightsTestNonPojo.java
```diff
@@ -0,0 +1,118 @@
...
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.json.JsonMapper;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.Objects;
+import java.util.UUID;
+
+public class FlightsTestNonPojo {
```
wchamber, line 16: worth a comment on what this is supposed to demonstrate.

src/main/java/bio/terra/stairway/FlightMap.java
jmczerk: When we met to discuss this, using binary buffers instead of `String` did come up, but we talked ourselves out of it, mainly due to the database implications.  We did reason that we could support binary formats by UTF-8 encoding with `String`.

And I agree that deprecation is out of scope for this PR.

src/main/java/bio/terra/stairway/DefaultFlightParameterDeserializer.java
```diff
@@ -0,0 +1,41 @@
...
+import java.io.IOException;
+
+/**
+ * This class implements default Jackson-based deserialization of classes and primitive types. It is
+ * intended for internal use by FlightMap and FlightFilter classes when a custom deserializer is not
+ * provided. It is not intended to be used directly.
+ *
+ * @param <T> Type that will be deserialized by this class.
+ */
+class DefaultFlightParameterDeserializer<T> implements FlightParameterDeserializer<T> {
```
wchamber, line 14: Is this 'default' as in it's a good default to use? Or is it the default when no serializer is specified, but we're trying to move clients away from it?
Can we configure an ObjectMapper so that it doesn't include the package name in the serialization? Is that a better 'default?'

wchamber:
> Regarding @wchamber suggestion of just putting/getting `String` instances, I’m not sure I agree, but I’m having a hard time coming up with a strong factual argument.
> 
> I feel like just using `String` is more likely to create disparate ad-hoc serdes logic, whereas the `FlightParameterSerializer`/`FlightParameterDeserializer` interfaces provide some degree of guardrails (if even just by making people think about things a little more intentionally), and are not onerous to implement against.
> 
> I think this is another decision we could revisit post merge, as I'm most concerned with propagating the database changes (which will support either convention).

I won't block the review, but I'm not especially persuaded. I don't really follow how the flight parameter interfaces make ad-hoc serdes less likely. Won't clients just implement ad-hoc implementations of the interfaces? We could try to propose a "good" default serdes scheme, but I don't know why we wouldn't want to bake it in then, or require special interfaces to go around it when clients can always create arbitrary strings.

Yes, developers need to be aware of the dangers of serialization across many binaries, but the interfaces to me don't seem to offer much support.

src/main/java/bio/terra/stairway/DefaultFlightParameterDeserializer.java
jmczerk: It's default as in "default when no serializer is specified".  We aspire to move clients away from it; the intent of this PR is to enable that, but not force the move (yet).

Not sure I follow the second question.  Do you mean configure an ObjectMapper for use with these Default classes which drops package name (as an enhancement to the "v2+" default formatting)?  If so, we could probably do this for the global Stairway ObjectMapper (as it only seems to be used for this purpose).

src/main/java/bio/terra/stairway/FlightMap.java
jmczerk: That moves the logic to the call site, which I believe is worse, as it leaks implementation details of the `FlightMap`.  The two non-test calls are in `FlightDao.java`.

src/test/java/bio/terra/stairway/fixtures/FlightsTestNonPojo.java
jmczerk: No; if we were to decide to go with a formal JSON-based approach to versioning, we'd probably want to wrap this boilerplate.  If we decided to go with, say, Protobuf, that would look different.

Some commenting/context on intention got lost here when I factored out the new test code to a separate PR.  If it's OK, I'd prefer to bring that in with that PR rather than try to cherry pick comments back.

src/test/java/bio/terra/stairway/fixtures/FlightsTestNonPojo.java
jmczerk: Per other comment, some commenting/context on intention got lost here when I factored out the new test code to a separate PR.  If it's OK, I'd prefer to bring that in with that PR rather than try to cherry pick comments back.

(the basic idea is to demo a toy class that had int numerator/denominator and refactored that to a float)

src/test/java/bio/terra/stairway/FlightMapTest.java
jmczerk: Main intent was to opportunistically generate test vectors for string literals in test cases.  OK to move this to DEBUG log level, or is the preference to remove?

src/main/java/bio/terra/stairway/FlightDao.java
jmczerk: Thinking about this, I think "parameters" bound this too much, and I think this should probably change to something like `DATABASE_SCHEMA_VERSION`, so that we have one constant to bump on schema updates.

src/main/java/bio/terra/stairway/FlightMap.java
jmczerk: I guess that comes down to what we’re worried about breaking.  I could see splitting this into two PR’s, one where the database schema gets updated and we write V1 (then propagate to modules), then another that starts writing V2.  It seems like that would be useful iff the database schema update part is what we are worried about; once we start writing V2 we’ll still have the same issue, unless we add another temporary phase where we write both (necessitating temporary columns for the “new” versions to live beside the old).  This is a lot of moving parts and seems likely to far offset the risk we’d be trying to mitigate.

All that said, the rolling upgrades of instances sharing a database seems like a **much** bigger issue.  We could probably solve it for this particular change using the “write both versions temporarily” approach above, but this becomes an issue again when versioning a stored value (which would need yet another approach).  Do we have a mechanism that we use for multiple instances sharing persistent data to notify when all instances have been upgraded?  

In clustered database and storage systems I’ve worked on in the past, the general practice for rolling upgrades was to use cluster-wide commit and rollback.  Nodes would write version N-1 formatted data on upgrade.  Upgrades would then be committed or rolled back at the cluster level; upon commit an event would be received that triggered writing version N formatted data.  I would imagine K8s provides the ability to know when all pods in a cluster are running version N (assuming database instance is shared within one cluster) and trigger a similar switchover; is there (industry standard or Terra) plumbing for this today?

PR #64, opened 2021-04-02
PF-316 provide a test fixture for checking FlightMap serialization

wchamber:
This is useful for helping clients determine what they can safely put in a FlightMap.
Example usage: https://github.com/DataBiosphere/terra-workspace-manager/pull/282

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,6 +10,7 @@ buildscript {
 
 plugins {
     id 'java-library'
+    id 'java-test-fixtures'
     id 'maven-publish'
     id 'com.jfrog.artifactory' version '4.13.0'
     id 'com.diffplug.gradle.spotless' version '3.27.2'
@@ -17,7 +18,8 @@ plugins {
 }
 
 group 'bio.terra'
-version '0.0.41-SNAPSHOT'
+// DO NOT SUBMIT
+version '0.1.0-SNAPSHOT'
 sourceCompatibility = 1.8
 
 // TODO: there may be a better way to supply these values than envvars.
@@ -37,6 +39,7 @@ gradle.taskGraph.whenReady { taskGraph ->
 
 repositories {
     mavenCentral()
+    mavenLocal()
 }
 
 dependencies {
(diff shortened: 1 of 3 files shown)


yonghaoy (approved):
The PR LG, not sure if you notice
One of my test case: `recreateFromSerializedStringError_finalClass` in
https://github.com/DataBiosphere/stairway/pull/63/files
is failed on another exception on different line: 
https://github.com/DataBiosphere/stairway/blob/develop/src/main/java/bio/terra/stairway/FlightMap.java#L56
But I think your the test fixturecover this. 
(I am sure that is new thing because my previous attemp: https://github.com/DataBiosphere/stairway/pull/49 passes that line, not sure what's chaning, probably java11 upgrade )

build.gradle
```diff
@@ -10,14 +10,16 @@ buildscript {
...
     id 'maven-publish'
     id 'com.jfrog.artifactory' version '4.13.0'
     id 'com.diffplug.gradle.spotless' version '3.27.2'
     id 'com.github.spotbugs' version '4.0.8'
 }
 
 group 'bio.terra'
-version '0.0.41-SNAPSHOT'
+// DO NOT SUBMIT
+version '0.1.0-SNAPSHOT'
```
zloery: Don't forget this - I assume the maven local bit is for testing without publishing?

zloery (approved):

build.gradle
wchamber: Thanks. Yep.

wchamber:
> The PR LG, not sure if you notice
> One of my test case: `recreateFromSerializedStringError_finalClass` in
> https://github.com/DataBiosphere/stairway/pull/63/files
> is failed on another exception on different line:
> https://github.com/DataBiosphere/stairway/blob/develop/src/main/java/bio/terra/stairway/FlightMap.java#L56
> But I think your the test fixturecover this.
> (I am sure that is new thing because my previous attemp: #49 passes that line, not sure what's chaning, probably java11 upgrade )

I'm not even getting an exception from the same line. Mine only throws locally deserializing on the ObjectMapper#readVAlue (FlightMap.java line 37).
Either way, at least this can be tested now, and we would catch it with the test fixture.

PR #66, opened 2021-04-05
PF-479 Fatally fail flights that cannot be recovered instead of aborting startup

wchamber:
Stairway can fail to deserialize the json of the FlightMap input and working maps. When this happens during flight recovery at start up time, Stairway fails to initialize. This can cause an entire service to be unable to start up if any Flight cannot be recovered. It's better to fatally fail the (hopefully few) unrecoverable flights than prevent the entier service from starting up.

Fatally failing the unrecoverable flights allows clients to move on, as they're not stuck polling for an operation that may never be resolved. This opts to fail fast rather than hope that a newer binary version will be able to handle the previously unrecoverable flight.

Added a test case that covers both the input map and the working map not being deserializable, which covers a couple of the examples from PF-479. The other example was explicitly addressed in PF-438.

### DIFF
--- a/src/main/java/bio/terra/stairway/FlightDao.java
+++ b/src/main/java/bio/terra/stairway/FlightDao.java
@@ -299,15 +299,17 @@ class FlightDao {
       startTransaction(connection);
       getStatement.setString("stairwayId", stairwayId);
       List<FlightContext> flightList = new ArrayList<>();
+      boolean flightsToDisown = false;
       try (ResultSet rs = getStatement.getPreparedStatement().executeQuery()) {
         while (rs.next()) {
+          flightsToDisown = true;
           String flightId = rs.getString("flightid");
           FlightContext flightContext;
           try {
             flightContext = makeFlightContext(connection, flightId, rs);
           } catch (JsonConversionException e) {
             logger.error(
-                String.format("Unable to make FlightContext to disown flight id %s", flightId));
+                "Unable to make FlightContext while recovering disowned flight id {}", flightId);
             // The flight will still be marked as READY by the database transaction, but the hook
             // will not be called.
             continue;
@@ -317,7 +319,7 @@ class FlightDao {
         }
       }
 
-      if (!flightList.isEmpty()) {
+      if (flightsToDisown) {
         updateStatement.setString("stairwayId", stairwayId);
         int disownCount = updateStatement.getPreparedStatement().executeUpdate();
         logger.info("Disowned " + disownCount + " flights for stairway: " + stairwayId);
--- a/src/main/java/bio/terra/stairway/FlightDao.java
+++ b/src/main/java/bio/terra/stairway/FlightDao.java
@@ -9,6 +9,7 @@ import bio.terra.stairway.exception.DuplicateFlightIdException;
 import bio.terra.stairway.exception.FlightException;
 import bio.terra.stairway.exception.FlightFilterException;
 import bio.terra.stairway.exception.FlightNotFoundException;
+import bio.terra.stairway.exception.JsonConversionException;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.sql.Connection;
@@ -300,7 +301,17 @@ class FlightDao {
       List<FlightContext> flightList = new ArrayList<>();
       try (ResultSet rs = getStatement.getPreparedStatement().executeQuery()) {
         while (rs.next()) {
-          FlightContext flightContext = makeFlightContext(connection, rs.getString("flightid"), rs);
+          String flightId = rs.getString("flightid");
+          FlightContext flightContext;
+          try {
+            flightContext = makeFlightContext(connection, flightId, rs);
+          } catch (JsonConversionException e) {
+            logger.error(
+                String.format("Unable to make FlightContext to disown flight id %s", flightId));
+            // The flight will still be marked as READY by the database transaction, but the hook
+            // will not be called.
+            continue;
+          }
           flightContext.setFlightStatus(FlightStatus.READY);
           flightList.add(flightContext);
         }
@@ -420,6 +431,66 @@ class FlightDao {
 
       commitTransaction(connection);
       hookWrapper.stateTransition(flightContext);
+    }
+  }
+
+  /**
+   * Fatally fail the flight id.
+   *
+   * <p>This should only be used if there are problems creating the FlightContext such that {@link
+   * #complete(FlightContext)} cannot be used.
+   *
+   * <p>This does not do a hook state transition as we do not have a FlightContext.
+   *
+   * @param flightId flight to fail fatally.
+   * @param fatalCause the reason to fail the flight fatally. May be null.
+   */
+  void fatalFail(String flightId, Exception fatalCause)
+      throws DatabaseOperationException, InterruptedException {
+    String serializedException;
+    try {
+      serializedException = exceptionSerializer.serialize(fatalCause);
+    } catch (Exception e) {
+      // We really do not want to throw while we're trying to fatally fail the flight.
+      logger.error("Unable to serialize fatalFail exception.", e);
+      serializedException = "Cause unknown. Unable to serialize fatalFail exception.";
+    }
+    String finalSerializedException = serializedException;
+    DbRetry.retryVoid("flight.fatalFail", () -> fatalFailInner(flightId, finalSerializedException));
+  }
+
+  private void fatalFailInner(String flightId, String serializedException) throws SQLException {
+    // Make this idempotent by not updating if the status is already FATAL.
+    final String sqlUpdateFlight =
+        "UPDATE "
+            + FLIGHT_TABLE
+            + " SET completed_time = CURRENT_TIMESTAMP,"
+            + " status = 'FATAL',"
+            + " serialized_exception = :serializedException,"
+            + " stairway_id = NULL"
+            + " WHERE flightid = :flightId AND status != 'FATAL'";
+
+    // The delete is harmless if it has been done before. We just won't find anything.
+    final String sqlDeleteFlightLog =
+        "DELETE FROM " + FLIGHT_LOG_TABLE + " WHERE flightid = :flightId";
+
+    try (Connection connection = dataSource.getConnection();
+        NamedParameterPreparedStatement statement =
+            new NamedParameterPreparedStatement(connection, sqlUpdateFlight);
+        NamedParameterPreparedStatement deleteStatement =
+            new NamedParameterPreparedStatement(connection, sqlDeleteFlightLog)) {
+
+      startTransaction(connection);
+      statement.setString("serializedException", serializedException);
+      statement.setString("flightId", flightId);
+      statement.getPreparedStatement().executeUpdate();
+
+      if (!keepFlightLog) {
+        deleteStatement.setString("flightId", flightId);
+        deleteStatement.getPreparedStatement().executeUpdate();
+      }
+
+      commitTransaction(connection);
     }
   }
 
@@ -571,7 +642,7 @@ class FlightDao {
   // The results set must be positioned at a row from the flight table,
   // and the select list must include debug_info, class_name, and status.
   private FlightContext makeFlightContext(Connection connection, String flightId, ResultSet rs)
-      throws InterruptedException, DatabaseOperationException, SQLException {
+      throws InterruptedException, JsonConversionException, SQLException {
     List<FlightInput> inputList = retrieveInputParameters(connection, flightId);
     FlightMap inputParameters = new FlightMap(inputList);
     FlightDebugInfo debugInfo = null;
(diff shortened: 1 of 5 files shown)


src/main/java/bio/terra/stairway/Stairway.java
```diff
@@ -897,11 +900,19 @@ void recoverReady()
...
+          queueFlight(flightContext);
+        } else {
+          resume(flightId);
+        }
+      } catch (MakeFlightException | JsonConversionException e) {
+        // Catch only exceptions that are unlikely to be resolved with retries. Allow the rest of
+        // recovery to continue.
+        logger.error(String.format("Unable to recover flight id %s", flightId), e);
+        // Mark the flight as having fatally failed so it does not stay un-recovered forever.
+        flightDao.fatalFail(flightId, e);
```
yonghaoy: `fatalFail` method may also throw exceptions, and that would also cause Stairway fail to initialize. 
Do you think we should initialize the flight when that happens or just leave it in RUNNING state, and hope next time recover call can pick it up?

yonghaoy (commented):
Overall looks good and reasonable to me. For the decision of state change: RUNNING-> FATAL. I defer to Dan's opinion.

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -300,7 +301,17 @@ private void disownRecoveryInner(String stairwayId)
...
       try (ResultSet rs = getStatement.getPreparedStatement().executeQuery()) {
         while (rs.next()) {
-          FlightContext flightContext = makeFlightContext(connection, rs.getString("flightid"), rs);
+          String flightId = rs.getString("flightid");
+          FlightContext flightContext;
+          try {
+            flightContext = makeFlightContext(connection, flightId, rs);
+          } catch (JsonConversionException e) {
+            logger.error(
+                String.format("Unable to make FlightContext to disown flight id %s", flightId));
```
jaycarlton: nit: you don't need String.format with the logger

The wording is confusing me. Are you disowning the flight because you can't make a context, or failing to disown it?

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -423,6 +434,66 @@ private void completeInner(FlightContext flightContext) throws SQLException {
...
+      throws DatabaseOperationException, InterruptedException {
+    String serializedException;
+    try {
+      serializedException = exceptionSerializer.serialize(fatalCause);
+    } catch (Exception e) {
+      // We really do not want to throw while we're trying to fatally fail the flight.
+      logger.error("Unable to serialize fatalFail exception.", e);
+      serializedException = "Cause unknown. Unable to serialize fatalFail exception.";
+    }
+    String finalSerializedException = serializedException;
```
jaycarlton, line 460: why do you need this reference to the string?

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -423,6 +434,66 @@ private void completeInner(FlightContext flightContext) throws SQLException {
...
+
+  private void fatalFailInner(String flightId, String serializedException) throws SQLException {
+    // Make this idempotent by not updating if the status is already FATAL.
+    final String sqlUpdateFlight =
+        "UPDATE "
+            + FLIGHT_TABLE
+            + " SET completed_time = CURRENT_TIMESTAMP,"
+            + " status = 'FATAL',"
+            + " serialized_exception = :serializedException,"
+            + " stairway_id = NULL"
```
jaycarlton, line 472: aside: the convention of setting the stairway id to null when we finish a flight seems like it would hamper debugging.

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -423,6 +434,66 @@ private void completeInner(FlightContext flightContext) throws SQLException {
...
+            new NamedParameterPreparedStatement(connection, sqlUpdateFlight);
+        NamedParameterPreparedStatement deleteStatement =
+            new NamedParameterPreparedStatement(connection, sqlDeleteFlightLog)) {
+
+      startTransaction(connection);
+      statement.setString("serializedException", serializedException);
+      statement.setString("flightId", flightId);
+      statement.getPreparedStatement().executeUpdate();
+
+      if (!keepFlightLog) {
```
jaycarlton, line 490: do we not normally keep the flight logs? I certainly see plenty of them in the database.

src/test/java/bio/terra/stairway/RecoveryTest.java
```diff
@@ -196,4 +199,84 @@ public void stopStepResultTest() throws Exception {
...
+    TestPauseController.setControl(1);
+    Stairway stairway2 =
+        Stairway.newBuilder()
+            .stairwayClusterName("stairway-cluster")
+            .stairwayName(stairwayName)
+            .workQueueProjectId(null)
+            .maxParallelFlights(3)
+            .build();
+    List<String> recordedStairways = stairway2.initialize(dataSource, false, false);
+    assertThat("One obsolete stairway to recover", recordedStairways.size(), equalTo(1));
```
jaycarlton: nit: is there a `hasSize` matcher you can use

src/main/java/bio/terra/stairway/Stairway.java
wchamber: Good point, added one last catch.
We already tried to initialize the flight, so we're pretty much out of options at this point and have to leave it in a running state. We can hope the next recover can pick it up, but we should really have alerting if this rare case starts to happen.

src/test/java/bio/terra/stairway/RecoveryTest.java
wchamber: done

src/main/java/bio/terra/stairway/FlightDao.java
wchamber: following what's done in `completeInner`. This is a parameter on Stairway.Builder that can be modified. I won't comment on what's the right default.

src/main/java/bio/terra/stairway/FlightDao.java
wchamber: Java lambdas, the string must be effectively final to be referenced in the retry lambda.

src/main/java/bio/terra/stairway/FlightDao.java
wchamber: done thanks. I get messed up about it not working when you want to format and log the exception....
Tried to clarify the wording, PTAL.

src/main/java/bio/terra/stairway/FlightDao.java
jaycarlton: ah

jaycarlton (approved):

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -571,7 +644,7 @@ private FlightContext makeFlightContextByIdInner(String flightId)
   // The results set must be positioned at a row from the flight table,
   // and the select list must include debug_info, class_name, and status.
   private FlightContext makeFlightContext(Connection connection, String flightId, ResultSet rs)
-      throws InterruptedException, DatabaseOperationException, SQLException {
+      throws InterruptedException, JsonConversionException, SQLException {
```
ddietterich, line 647: JsonConversionException is not a checked exception.

ddietterich (changes requested):
I think it is worth discussing whether it is always the right thing to allow a service to come up if it cannot recover in-progress flights. If the service cannot recover, that says the service has a bug. It maybe had the bug when it wrote the data to Stairway because it didn't test serdes, or it may be an improper versioning bug, but there is a service bug.

Abandoning a flight means needing to perform manual intervention to undo/fix whatever the flight was trying to do. Otherwise, we abandon resources, block progress due to partially created objects, corrupt application database state, etc. I am more inclined to add an application that allows us to force-fail a flight in the database as a manual step, rather than code that as the recovery behavior.

In this specific code, I think it is incorrect to set a flight to READY state when you cannot construct a flight context. It would be better to set that flight directly to a FATAL state _in that transaction_. Otherwise, a failure can intervene between the disown transaction and the later time when you transition to fatal. That appears to leave an orphan READY flight that is not, in fact, runnable.

We could consider a state other than FATAL that would allow for recovery using a fixed version of the service code.

wchamber:
> I think it is worth discussing whether it is always the right thing to allow a service to come up if it cannot recover in-progress flights. If the service cannot recover, that says the service has a bug. It maybe had the bug when it wrote the data to Stairway because it didn't test serdes, or it may be an improper versioning bug, but there is a service bug.
> 
> Abandoning a flight means needing to perform manual intervention to undo/fix whatever the flight was trying to do. Otherwise, we abandon resources, block progress due to partially created objects, corrupt application database state, etc. I am more inclined to add an application that allows us to force-fail a flight in the database as a manual step, rather than code that as the recovery behavior.
> 
> In this specific code, I think it is incorrect to set a flight to READY state when you cannot construct a flight context. It would be better to set that flight directly to a FATAL state _in that transaction_. Otherwise, a failure can intervene between the disown transaction and the later time when you transition to fatal. That appears to leave an orphan READY flight that is not, in fact, runnable.
> 
> We could consider a state other than FATAL that would allow for recovery using a fixed version of the service code.

@ddietterich  Let's try an async round of discussion, then probably worth having a meeting.

I don't think it's right to allow a single bad flight to bring down the entire service. Yes, the service in some form has a bug, but it is it the kind of bug that we want to default to total unavailability, or need to manually recover the set of flights that were bugged. Forbidding the entire service from operating until there's some manual intervention seems like a good way to self-inflict outages. Needing to manually recover bugged flights seems like the consequence of a run-of-the-mill kind of bug.
Worth also considering what this means for clients - should they poll indefinitely on a flight, or should we let them know their flight has failed.

Agreed on not going through te intermediate READY state. If we want to proceed, I'll revisit.
Another state other than FATAL that keeps more of the progress of the flight sounds good to me.

wchamber:
We can separate the "fail the flight" from "allow the service to start up," e.g. put the flight in some pending UNWORKABLE state, but don't mark it as complete. Allow the service to come up. Then the manual intervention can still resume the bugged flights, while the rest of the service can make progress.
I think the fail vs don't fail the bugged flight looks like a consideration of what a client might want and how we want to resolve flights that have failed and are therefore not atomic.

ddietterich:
> > I think it is worth discussing whether it is always the right thing to allow a service to come up if it cannot recover in-progress flights. If the service cannot recover, that says the service has a bug. It maybe had the bug when it wrote the data to Stairway because it didn't test serdes, or it may be an improper versioning bug, but there is a service bug.
> > Abandoning a flight means needing to perform manual intervention to undo/fix whatever the flight was trying to do. Otherwise, we abandon resources, block progress due to partially created objects, corrupt application database state, etc. I am more inclined to add an application that allows us to force-fail a flight in the database as a manual step, rather than code that as the recovery behavior.
> > In this specific code, I think it is incorrect to set a flight to READY state when you cannot construct a flight context. It would be better to set that flight directly to a FATAL state _in that transaction_. Otherwise, a failure can intervene between the disown transaction and the later time when you transition to fatal. That appears to leave an orphan READY flight that is not, in fact, runnable.
> > We could consider a state other than FATAL that would allow for recovery using a fixed version of the service code.
> 
> @ddietterich Let's try an async round of discussion, then probably worth having a meeting.
> 
> I don't think it's right to allow a single bad flight to bring down the entire service. Yes, the service in some form has a bug, but it is it the kind of bug that we want to default to total unavailability, or need to manually recover the set of flights that were bugged. Forbidding the entire service from operating until there's some manual intervention seems like a good way to self-inflict outages. Needing to manually recover bugged flights seems like the consequence of a run-of-the-mill kind of bug.
> Worth also considering what this means for clients - should they poll indefinitely on a flight, or should we let them know their flight has failed.
> 
> Agreed on not going through te intermediate READY state. If we want to proceed, I'll revisit.
> Another state other than FATAL that keeps more of the progress of the flight sounds good to me.

I think there are two types of failure here:
1. Version X wrote data that version X cannot read back
2. Version X wrote data that version X+1 cannot read back

For case 1, there are two ways out. One is forcing failure of the unreadable flights. The other is to implement version X+1 that can read the flights and deploy it.

For case 2, version X+1 should not be deployed. It is broken. Not upward compatible. It should not run. It should not be allowed to trash perfectly good flights.

My suggestion is that we make a tool that will allow us to manually force-fail flights. That allows us to surgically address case 1 without allowing case 2. My argument is that dealing with the force-fail flights requires manual intervention anyway, so it is not a significant burden to make the force-fail operation manual instead of automatic.

We added a debug mode in Stairway to allow testing for case 1. It is possible to run the entire regression suite of any service with that enabled and ensure that everything written is readable. It is up to us to apply that testing regimen and reduce the chances of case 1 going forward.

wchamber:
We sync'd offline. Trying to summarize here:
The main tradeoff is around what failure case is more frequent * painful: Refusing to start up the service if flights can't be recovered may help us prevent bad binaries going out to production that write flights that can't be recovered. OTOH, we may get an outage in production if we still get a bad flight and on't manually fix it quickly enough. We're going to wait and see which, if any, problem we see more frequently before adding more code.
We agreed in the meanwhile to look into with more priority adding more admin endpoints on both operating and down services for modifying flights. Operations such as asking a flight to reverse direction and error, or fatally fail a flight.

PR #68, opened 2021-04-15
PF-694 Add FlightDebugInfo support for failing after the last step.

wchamber:
Extend the FlightDebugInfo in Stairway to allow it to be easy to test undoing a flight by setting a failure after the last do step executes.

By adding a field, existing versions of Stairway will not be able to deserialize FlightDebugInfo's serialized by this and newer versions going forward: they will be [unable to make a FlightContext](https://github.com/DataBiosphere/stairway/blob/cdaed8050bc3a4c1729f1d7d6f9bcd9987307c76/src/main/java/bio/terra/stairway/FlightDao.java#L578-L586). Older versions serializations will be able to be deserialized by this version, as ObjectMapper's FAIL_ON_UNKNOWN_PROPERTIES defaults to false. this is a backwards incompatible change for FlightDebugInfo serialization.
Set `DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES` to make this not the case in the future.

### DIFF
--- a/src/main/java/bio/terra/stairway/Flight.java
+++ b/src/main/java/bio/terra/stairway/Flight.java
@@ -278,33 +278,10 @@ public class Flight implements Runnable {
 
         if (context().isDoing()) {
           result = currentStep.step.doStep(context());
-
-          // If we are in debug mode and a failure is set for this step AND we have not already
-          // failed here, then insert a failure. We do this right after the step completes but
-          // before the flight logs it so that we can look for dangerous UNDOs.
-          if (context().getDebugInfo() != null
-              && context().getDebugInfo().getFailAtSteps() != null
-              && context().isDoing()
-              && context().getDebugInfo().getFailAtSteps().containsKey(context().getStepIndex())
-              && !debugStepsFailed.contains(context().getStepIndex())) {
-            StepStatus failStatus =
-                this.context().getDebugInfo().getFailAtSteps().get(context().getStepIndex());
-            logger.info(
-                "Failed for debug mode fail step at step {} with result {}",
-                context().getStepIndex(),
-                failStatus);
-            result = new StepResult(failStatus);
-            debugStepsFailed.add(context().getStepIndex());
-          } else if (context().getDebugInfo() != null
-              && context().getDebugInfo().getLastStepFailure()
-              && context().getStepIndex() == steps.size() - 1) {
-            // If we are in debug mode for failing at the last step and this is the last step,
-            // insert a failure.
-            logger.info("Failed for debug mode last step failure.");
-            result = new StepResult(StepStatus.STEP_RESULT_FAILURE_FATAL);
-          }
+          result = debugStatusReplacement(result);
         } else {
           result = currentStep.step.undoStep(context());
+          result = debugStatusReplacement(result);
         }
       } catch (InterruptedException ex) {
         // Interrupted exception - we assume this means that the thread pool is shutting down and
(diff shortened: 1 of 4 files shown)


src/main/java/bio/terra/stairway/FlightDebugInfo.java
```diff
@@ -12,6 +12,9 @@
 
   private static final ObjectMapper objectMapper = new ObjectMapper();
```
ddietterich: Suggest you add
```
.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
```
That will allow us to add optional properties and have them ignored by previous versions. As it should have been...

src/main/java/bio/terra/stairway/Flight.java
```diff
@@ -287,10 +287,21 @@ private StepResult stepWithRetry() throws InterruptedException, StairwayExecutio
               && context().isDoing()
               && context().getDebugInfo().getFailAtSteps().containsKey(context().getStepIndex())
               && !debugStepsFailed.contains(context().getStepIndex())) {
-            result =
-                new StepResult(
-                    this.context().getDebugInfo().getFailAtSteps().get(context().getStepIndex()));
+            StepStatus failStatus =
```
ddietterich: This clause is getting big and is already pretty nested. How about putting in a separate method. Something like:
```
StepResult debugStatusReplacement(FlightContext context, StepResult result)
```
And then if there is no debug info, we return the input result. If there is, we substitute the replacement.

What do you think?

ddietterich (approved):
Looks good. Two suggestions for you

src/main/java/bio/terra/stairway/FlightDebugInfo.java
wchamber: Done.
I didn't realize that FAIL_ON_UNKNOWN_PRIMITIVES was default false. So now I think v0 serialized FlightDebugInfo will be readable by this new v1 reader, but v1 written FlightDebugInfo will not be readable by v0 readers due to FAIL_ON_UNKNOWN_PROPERTIES. Updating PR description accordingly.
I think this is ok to roll out because we don't use FlightDebugInfo outside of testing, right?

src/main/java/bio/terra/stairway/Flight.java
wchamber: SGTM. I went ahead and started calling it on the undo case, even though both of our current status replacements are only for the DO direction. That ok too, or should I leave it as only being called on the DO result (w/ or w/out the then unnecessary context().isDoing() checking)?

src/main/java/bio/terra/stairway/Flight.java
```diff
@@ -348,6 +336,43 @@ private StepResult stepWithRetry() throws InterruptedException, StairwayExecutio
...
+          failStatus);
+      debugStepsFailed.add(context().getStepIndex());
+      return new StepResult(failStatus);
+    }
+    // If we are in debug mode for failing at the last step, and this is the last step, insert a
+    // failure.
+    if (debugInfo.getLastStepFailure()
+        && context().isDoing()
+        && context().getStepIndex() == steps.size() - 1) {
+      logger.info("Failed for debug mode last step failure.");
```
tnadler, line 366: For consistency, probably worth adding this step index to debugStepsFailed from above. Shouldn't have an effect now, but in case we ever use that field for something else.

tnadler (approved):
One quick thing but otherwise looks good.

src/main/java/bio/terra/stairway/Flight.java
wchamber: Hmmm. As commented, it's specific to failAtSteps and I think it's a little more confusing to explain its dual purpose when that dual purpose isn't relied on. I'm a little inclined to leave as-is and YAGNI let future changes change the field as it's useful for them. Could be persuaded.

src/main/java/bio/terra/stairway/FlightDebugInfo.java
ddietterich: Right on both counts.

src/main/java/bio/terra/stairway/Flight.java
ddietterich: We do not want to inject failures in the undo, because that just induces DISMAL FAILURE results.

ddietterich (approved):
LGTM

src/main/java/bio/terra/stairway/Flight.java
wchamber: Done, thanks.

PR #71, opened 2021-04-20
[PF-580] Stairway Serialization Phase One

jmczerk:
This PR represents Phase 1 of the Serialization as discussed [in this document](https://docs.google.com/document/d/1b1VckTBb_V1pCMhoN2GgHnGDR2f8MV2HuwIgDqFc_9M):

* ~`FlightMap` maps to `String` instances containing serialized data; data is serialized at `put()` time and deserialized at `get()` time.~
* ~`FlightMap` gets `rawGet()` and `rawPut()` methods for getting/putting raw (serialized) data.~
* Begin to phase out `flightlog::working_parameters` and `flight::output_parameters` JSON Map columns in favor of new `flightworking` table.
* The `keepFlightLog` parameter is now ignored and flight logs are always kept (all known clients were getting this default behavior already).


src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -95,14 +97,15 @@ private void submitInner(FlightContext flightContext)
         "INSERT INTO "
             + FLIGHT_TABLE
             + " (flightId, submit_time, class_name, status, stairway_id, debug_info)"
-            + "VALUES (:flightId, CURRENT_TIMESTAMP, :className, :status, :stairwayId, :debugInfo)";
```
jmczerk: One interesting learning here: if we're going to generate the timestamp on the host anywhere (which we do elsewhere so that we can keep it in sync between `flightlog|flight` and `flightworking`), we should do it everywhere.

I discovered this because it's apparently possible to have clock skew between the Linux VM that Docker runs containers on in MacOS and the MacOS system clock, which lead to unit test scenarios where `submit_time` got ahead of `completed_time` (which obviously broke test invariants).

jmczerk:
One could probably make a case for renaming `FlightInput` to `FlightParameter`, given it's used to represent "working" and "output" parameters now too.  I suspect that would add a lot of noise to the PR though; interested in folks' thoughts.

src/main/java/bio/terra/stairway/FlightDao.java
ddietterich: The other alternative would have been to use returned values from the first insert to get the timestamp to use  on the later insert. That might be better, because Postgres will give us a consistent clock.

I don't know how consistent the clocks are on different pods, but we could see an apparent ordering in the database that was not the actual ordering. Although, I'm hard pressed to think of a case where that matters.

Either way, this deserves a comment at the top of the class about handling timestamps - either all in or all out.

src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -77,33 +102,78 @@ public void makeImmutable() {
...
+   * Stores a raw (serialized) data value for a given key.
+   *
+   * @param key Key to store the data under.
+   * @param value Data to store.
+   */
+  public void putRaw(String key, String value) {
     map.put(key, value);
   }
 
   public String toJson() {
```
ddietterich, line 169: Should this be `@Deprecated` now? And `fromJson` also? We won't be converting the whole map to and from json once the transition is complete, right?

src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -77,33 +102,78 @@ public void makeImmutable() {
...
     try {
-      return getObjectMapper().writeValueAsString(map);
-    } catch (JsonProcessingException ex) {
+      // Serializing as Map<String, RawJsonValue> will produce a format compatible with older code's
+      // Map<String, Object> representation.
+      Map<String, RawJsonValue> temporaryMap = new HashMap<>();
+      for (Map.Entry<String, String> entry : map.entrySet()) {
+        temporaryMap.put(entry.getKey(), RawJsonValue.create(entry.getValue()));
```
ddietterich: Why not just perform the serialization of the object here?
It looks to me like going through the RawJsonType and custom serializer stuff does not get a different result, but spreads the logic out a lot making it harder to follow.

src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -77,33 +102,78 @@ public void makeImmutable() {
...
       throw new JsonConversionException("Failed to convert map to json string", ex);
     }
   }
 
   public void fromJson(String json) {
     try {
-      map = getObjectMapper().readValue(json, new TypeReference<Map<String, Object>>() {});
+      // Deserialize Map<String, Object> written by previous versions of code; deserializing as
+      // Map<String, JsonNode> allows us to defer deserialization and obtain String representations
+      // to store in our Map<String, String>.
```
ddietterich: I'm curious about the interaction of this readValue with the embedded typing in the StairwayMapper.
Do we need to use a mapper that will not process the embedded types?

src/test/java/bio/terra/stairway/FlightMapTest.java
```diff
@@ -0,0 +1,57 @@
...
+    Assertions.assertEquals(intIn, intOut);
+  }
+
+  @Test
+  public void fromJsonMap() {
+    FlightMap map = new FlightMap();
+
+    String jsonMap =
+        "[\"java.util.HashMap\",{\"mykey\":[\"bio.terra.stairway.fixtures.FlightsTestPojo\",{\"astring\":\"mystring\",\"anint\":99}]}]";
+
```
ddietterich, line 118: Maybe we could harvest a bunch of samples from the TDR dev stairway database and run them through this logic to see if we get any surprises.

ddietterich (commented):
I don't understand how some of this works.

src/main/java/bio/terra/stairway/FlightContext.java
```diff
@@ -104,6 +105,16 @@ public FlightMap getWorkingMap() {
...
   }
 
+  /**
+   * Replace the default map created at constructor time.
+   *
+   * @param map If not empty, replaces the existing workingMap with the contained map. If empty,
+   *     existing map is left as-is.
+   */
+  public void setWorkingMap(Optional<FlightMap> map) {
+    if (map.isPresent()) workingMap = map.get();
```
wchamber: nit: always use brackets. https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -164,15 +170,25 @@ private void stepInner(FlightContext flightContext) throws SQLException {
             new NamedParameterPreparedStatement(connection, sqlInsertFlightLog)) {
       startTransaction(connection);
       statement.setString("flightId", flightContext.getFlightId());
+      statement.setInstant("logTime", logTime);
+
+      // TODO: Column working_parameters is being phased out in favor of table flightworking.
```
wchamber: nit: prefer to TODO(PF-NNN) a ticket for cleanup. Makes searching later easier. Ditto others

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -841,6 +891,30 @@ private void storeInputParameters(
...
 
+  private void storeWorkingParameters(
+      Connection connection, String flightId, Instant logTime, FlightMap workingParameters)
+      throws SQLException {
+    List<FlightInput> inputList = workingParameters.makeFlightInputList();
+
+    final String sqlInsertInput =
+        "INSERT INTO "
+            + FLIGHT_WORKING_TABLE
+            + " (flightId, log_time, key, value) VALUES (:flightId, :logTime, :key, :value)";
```
wchamber: Is the log_time in FLIGHT_WORKING_TABLE acting as a foreign key for the FLIGHT_LOG_TABLE? This feels sneaky and surprising. Should we just use a unique FLIGHT_LOG_TABLE key as a foreign key instead?

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -396,28 +415,44 @@ private void completeInner(FlightContext flightContext) throws SQLException {
...
+      //       We continue to write the output map here temporarily for backward compatibility.
       statement.setString("outputParameters", flightContext.getWorkingMap().toJson());
+
+      statement.setInstant("completedTime", completedTime);
       statement.setString("status", flightContext.getFlightStatus().name());
       statement.setString("serializedException", serializedException);
       statement.setString("flightId", flightContext.getFlightId());
       statement.getPreparedStatement().executeUpdate();
 
       if (!keepFlightLog) {
```
wchamber: probably worth updating the documentation about this field to mention the working map table.

src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -5,21 +5,24 @@
...
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 
 /**
  * FlightMap wraps a {@code HashMap<String, Object>} It provides a subset of the HashMap methods. It
  * localizes code that casts from Object to the target type. It provides a way to set the map to be
  * immutable.
```
wchamber: update me

src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -77,33 +102,78 @@ public void makeImmutable() {
...
     map.put(key, value);
   }
 
   public String toJson() {
     try {
-      return getObjectMapper().writeValueAsString(map);
-    } catch (JsonProcessingException ex) {
+      // Serializing as Map<String, RawJsonValue> will produce a format compatible with older code's
+      // Map<String, Object> representation.
+      Map<String, RawJsonValue> temporaryMap = new HashMap<>();
```
wchamber: Not if the older code doesn't have the RawJsonValue class.If we're trying to prevent rollback/concurrent version problems, this doesn't work. 
Consider what's at head as version N and this PR merged as version N + 1.  version N can't read what version N + 1 wrote, as it won't know what a `RawJsonValue` is.
Am I misunderstanding how the custom serializer will write this out?

I thought the intention was to allow this PR to be rolled back if necessary. We need to write just in the old version format to do that. Unless you want to do a preemptory PR introducing `RawJsonValue`.

wchamber: Or if it does create something that's backwards compatible with version N, why bother with this RawJsonValue and just use the version N serialization?

jmczerk: The reason for mapping to `RawJsonValue` class is that we can then inject a custom serializer that gets hooked for type `RawJsonValue`, which injects the raw JSON string (including the object type).  That's the highest layer in Jackson where we can hook raw JSON; upper layers will escape it and treat it as a string, lower layers require us to roll the entire map.

src/main/java/bio/terra/stairway/FlightDao.java
ddietterich: I like that idea.

src/main/java/bio/terra/stairway/FlightMap.java
jmczerk: What this does is deserializes each value in the map to the root node of a Jackson tree, which is either a string/int for primitives ("value node") or an array ("array node") for types (entry 0 is a string representing the type, entry 1 is the payload as a JSON object.

Calling `toString()` on the root node renders the JSON for the tree, which is what we store in our `Map<String, String>`.

This is the closest thing to official docs I could find on Jackson tree model: https://www.baeldung.com/jackson-json-node-tree-model

src/main/java/bio/terra/stairway/FlightDao.java
jmczerk: Sure, this raises a question with using this table for output parameters though. If we clean up the logs we will lose the last set of working parameters which we now look at for output parameters.  Is that acceptable?

src/main/java/bio/terra/stairway/FlightMap.java
jmczerk: To this question and @wchamber question above, I was going to some lengths to use raw JSON semantics to enforce the restriction that we only ever ser/des **`Objects`** at get/put time, and traffic in JSON everywhere else.

If we are OK with relaxing that I can see if this can be simplified by deserializing and re-serializing into a `Map<String, Object>` and getting the same result.

The spreading out of the logic is due to not wanting to re-register the same serialization hooks with the static global ObjectMapper on every call, thus making it available to the global static instance creation.

src/main/java/bio/terra/stairway/FlightDao.java
jmczerk: I'd prefer to use `CURRENT_TIMESTAMP` and generate it in the database.  I did it like this as a first pass to avoid plumbing [this overload](https://docs.oracle.com/javase/7/docs/api/java/sql/Connection.html#prepareStatement(java.lang.String,%20java.lang.String[]))  through `NamedParameterPreparedStatement` without some discussion.  But it sounds like if we go with a unique ID key in `flightlog` this might be moot...

src/main/resources/stairway/db/changesets/20210416_flightworking.yaml
```diff
@@ -0,0 +1,49 @@
...
+            tableName: flightlog
+            columnName: working_parameters
+            remarks: deprecating for flightworking table
+        - addColumn:
+            tableName: flightlog
+            columns:
+              - column:
+                  name: id
+                  type: uuid
+                  constraints:
```
jmczerk, line 16: Postrgres cannot generate default UUID's without a plugin, so we cannot upgrade with a non-nullable constraint.  I did run all tests with a non-null constraint in place to try to catch non-null inserts in code.

src/main/resources/stairway/db/changesets/20210416_flightworking.yaml
```diff
@@ -0,0 +1,49 @@
...
+                    nullable: false
+        - addPrimaryKey:
+            tableName: flightworking
+            columnNames: flightlog_id, key
+            constraintName: pk_flightworking
+        - addForeignKeyConstraint:
+            baseTableName: flightworking
+            baseColumnNames: flightlog_id
+            referencedTableName: flightlog
+            referencedColumnNames: id
```
jmczerk, line 45: Interested in thoughts on cascaded deletes (I wrote delete logic in `FlightDao::deleteInner()` to handle this).

src/main/java/bio/terra/stairway/NamedParameterPreparedStatement.java
```diff
@@ -99,4 +100,8 @@ public void setString(String name, String value) throws SQLException {
   public void setInstant(String name, Instant value) throws SQLException {
     preparedStatement.setTimestamp(getIndex(name), Timestamp.from(value));
   }
+
```
jmczerk, line 103: Adding a `setObject()` felt a little wild-west, so I restricted it to UUID.  I'd be fine with making it more general if others prefer.

src/main/java/bio/terra/stairway/FlightDao.java
jmczerk: Restored original behavior.

src/main/java/bio/terra/stairway/FlightMap.java
jmczerk: Map<String, String> change needs to wait for the database changes to propagate due to the need to write backward compatible JSON maps.  Reverted most of FlightMap.

src/main/java/bio/terra/stairway/FlightMap.java
jmczerk: Reverted per previous comment.

src/main/java/bio/terra/stairway/FlightMap.java
jmczerk: Reverted per previous comment.  Next phase will remove toJson().

src/main/java/bio/terra/stairway/FlightMap.java
jmczerk: toJson() will go away sooner, fromJson() will need to stick around until all of the Flights written before this merges are gone.

src/test/java/bio/terra/stairway/FlightMapTest.java
jmczerk: I like that idea, though test will need to be written in TDR (and might be less useful now that we've reverted most of FlightMap).

src/main/java/bio/terra/stairway/FlightDao.java
jmczerk: We discussed just using flightworking for output parameters and removing the code that deletes FlightLog entries, so the last flightworking will always be available.

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -392,32 +407,25 @@ private void completeInner(FlightContext flightContext) throws SQLException {
...
 
+      // TODO(PF-703): Column output_parameters is being phased out in favor of table
+      //               flightworking.  We continue to write the output map here
+      //               temporarily for backward compatibility.
       statement.setString("outputParameters", flightContext.getWorkingMap().toJson());
+
       statement.setString("status", flightContext.getFlightStatus().name());
       statement.setString("serializedException", serializedException);
       statement.setString("flightId", flightContext.getFlightId());
       statement.getPreparedStatement().executeUpdate();
```
jmczerk, line 427: Per our discussion, `keepFlightLog` is no longer obeyed, and logs are always kept (this avoids FK violations and allows us to use last `flightworking` as `output_parameters`.

src/main/java/bio/terra/stairway/FlightDao.java
jmczerk: Went with unique ID in flight log instead.

src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -61,6 +66,57 @@ public FlightMap() {
...
+      return Optional.empty();
+    }
+
+    // TODO(PF-703): Once we stop writing JSON, we may still have JSON-only flight data in the
+    //               database, as well as flights with both. At that point we should favor
+    //               inputList over json.
+
+    FlightMap map = new FlightMap();
+    map.fromJson(json);
+
```
jmczerk, line 93: If we got a non-empty `inputList` from `flightworking`, the block below will validate that all of the values in `inputList` exist in the deserialized-from-JSON `FlightMap` and can be deserialized as the same type.  It does not fail if either check fails, but logs ERRORs.  Am looking for feedback on whether folks are OK with this as a way to build confidence before the next phase (where we stop writing JSON map to `flightlog::working_parameters`).

src/main/java/bio/terra/stairway/FlightContext.java
```diff
@@ -104,6 +105,18 @@ public FlightMap getWorkingMap() {
     return workingMap;
   }
 
+  /**
+   * Replace the default map created at constructor time.
+   *
+   * @param map If not empty, replaces the existing workingMap with the contained map. If empty,
+   *     existing map is left as-is.
+   */
+  public void setWorkingMap(Optional<FlightMap> map) {
```
wchamber: tiny nit: taking an optional as an argument is a little funny to me here. Maybe just take `FlightMap` and force the client to only call the function if they have something to set?

src/main/java/bio/terra/stairway/FlightDao.java
wchamber: worth an update to the PR description

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -637,13 +657,23 @@ private void fillFlightContexts(Connection connection, List<FlightContext> fligh
...
             flightContext.setRerun(rsflight.getBoolean("rerun"));
             flightContext.setDirection(Direction.valueOf(rsflight.getString("direction")));
             flightContext.setResult(stepResult);
             flightContext.setFlightStatus(FlightStatus.valueOf(rsflight.getString("status")));
             flightContext.setStepIndex(rsflight.getInt("step_index"));
+
+            // TODO(PF-703): In the current transition away from working_parameters column towards
+            //               flightworking table we can either have only JSON or both.  Until we've
+            //               transitioned, delegate the decision of which to use to the FlightMap
+            //               class.
```
wchamber: whitespace nit: I don't think we're treating TODOs as indenting comments in any other repo. nit for consistency, ditto others.

src/main/java/bio/terra/stairway/FlightMap.java
ddietterich: It looks reasonable to me.

src/main/java/bio/terra/stairway/FlightState.java
```diff
@@ -11,7 +11,7 @@
   private FlightMap inputParameters;
   private Instant submitted;
   private Instant completed;
-  private FlightMap resultMap; // filled in when flightStatus is SUCCESS
+  private Optional<FlightMap> resultMap; // filled in when flightStatus is SUCCESS
```
ddietterich: I think this is not upward compatible. Isn't this object returned as the output of several Stairway methods?

src/main/java/bio/terra/stairway/Stairway.java
```diff
@@ -161,7 +161,7 @@ public Builder enableWorkQueue(boolean enableWorkQueue) {
      * flightlog table at the completion of a flight. This flag can be set on Stairway create to
      * request keeping the flightlog instead of cleaning it.
      *
-     * @param keepFlightLog true to keep the flight log. Default is false;
+     * @param keepFlightLog true to keep the flight log. Default is true;
```
ddietterich, line 164: Maybe `@Deprecate` the method?

src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -61,6 +66,74 @@ public FlightMap() {
...
+    // TODO(PF-703): Remove this block.
+    if (!inputList.isEmpty()) {
+      try {
+        map.validateAgainst(inputList);
+      } catch (final Exception ex) {
+        Logger logger = LoggerFactory.getLogger("FlightMap");
```
wchamber: nits: No need for Exception to be `final`. prefer a single class logger.

src/main/java/bio/terra/stairway/FlightState.java
```diff
@@ -60,12 +60,12 @@ public void setCompleted(Instant completed) {
...
 
   public Optional<FlightMap> getResultMap() {
-    return Optional.ofNullable(resultMap);
+    return resultMap;
   }
 
-  public void setResultMap(FlightMap resultMap) {
-    if (resultMap != null) {
-      resultMap.makeImmutable();
+  public void setResultMap(Optional<FlightMap> resultMap) {
```
wchamber: nit: if we keep the no-op empty Optional set on `FlightContext`, let's add a comment here that this always overrides the result map. Otherwise, this doesn't need a comment.

src/main/java/bio/terra/stairway/Stairway.java
wchamber, line 164: is this function deprecated/non functional now?

src/main/resources/stairway/db/changesets/20210416_flightworking.yaml
```diff
@@ -0,0 +1,49 @@
...
+        - addColumn:
+            tableName: flightlog
+            columns:
+              - column:
+                  name: id
+                  type: uuid
+                  constraints:
+                    unique: true
+        - createTable:
+            tableName: flightworking
```
wchamber, line 19: optional: maybe `flightworkingparameters` or `flightworkingmap`? `flightworking` feels like it could be almost anything.

src/main/resources/stairway/db/changesets/20210416_flightworking.yaml
wchamber: I don't have any objections to cascaded deletes, but I also don't have much practical experience with them.

wchamber (approved):
lgtm, few nitpicks.

src/main/resources/stairway/db/changesets/20210416_flightworking.yaml
ddietterich: Either way works for me. Given that we are asking the DB to do the constraint checking, it makes sense to me to let the DB do the cascading as well.

src/test/java/bio/terra/stairway/FlightMapTest.java
ddietterich: Yes, true. It wouldn't work. I think this is fine.

ddietterich (changes requested):
I think there is one incompatible thing that needs to be fixed.
Otherwise, looks good.

src/main/java/bio/terra/stairway/FlightState.java
ddietterich: Ah, it doesn't matter for compatibility, but still not a fan of Optional used as other than an output.

src/main/java/bio/terra/stairway/FlightContext.java
jmczerk: Reverted to existing code and did check external to FlightContext.

src/main/java/bio/terra/stairway/FlightState.java
jmczerk: Reverted to existing code, null checking moved out of FlightState.

src/main/java/bio/terra/stairway/FlightState.java
jmczerk: Reverted out Optional to existing code, did null checking externally to FlightState.

src/main/resources/stairway/db/changesets/20210416_flightworking.yaml
jmczerk: Unless @ddietterich has a preference for something else, I'll stick with his original naming.

ddietterich: I think it is fine the way it is, FWIW

ddietterich (approved):
LGTM

PR #72, opened 2021-04-20
PF-699 Add FlightDebugInfo support for failing do & undo by Step class name

wchamber:
Extend Stairway's FlightDebugInfo to allow changing the status of step's by the Step's class name instead of a step index. Step index could be changed by adding more steps to a Flight, but by using the Step.class class name, it's easier to express what's being tested and fail to compile if a test's step name changes.

Allow the setting failures on both `do` and `undo` for a Step, as it's important to verify that undo operations are also idempotent to avoid dismal failures.

This will not support the same step class name being run multiple times, bu that does not seem to be as useful.

### DIFF
--- a/src/main/java/bio/terra/stairway/Flight.java
+++ b/src/main/java/bio/terra/stairway/Flight.java
@@ -42,9 +42,12 @@ public class Flight implements Runnable {
   private FlightContext flightContext;
   private final Object applicationContext;
 
-  // This list will only be populated if debugInfo.FailAtSteps is populated. If so, this
-  // keeps track of which steps have already been failed so we do not infinitely retry.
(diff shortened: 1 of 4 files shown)


ddietterich (approved):
LGTM, thanks

ddietterich:
I had voiced a concern about not wanting to fault-insert on undo because you'd cause dismal failures. I was wrong about that and glad that you provided both do and undo.

It is true, I think, that if you insert a STEP_FAILURE_FATAL it will stop the flight with a dismal failure. But other StepStatus work fine and can validate idempotency of undo steps.

tnadler (approved):
This is great, thank you

PR #73, opened 2021-04-21
PF-316 Revert providing a test fixture for checking FlightMap serialization

wchamber:
Revert PR #64. The test fixtures plugin is causing issues with Spring Boot, see https://github.com/gradle/gradle/issues/11696
We can worry about serialization checking a different way.


ddietterich (approved):
Thank you

jmczerk (approved):

PR #74, opened 2021-05-06
[PF-703] Stop writing to flightlog::working_parameters and flight::ou…

jmczerk:
- This PR corresponds to [Phase 2 as described in this document](https://docs.google.com/document/d/1b1VckTBb_V1pCMhoN2GgHnGDR2f8MV2HuwIgDqFc_9M/edit#heading=h.49331srrvaui):
  - Stop writing JSON maps to `flightlog::working_parameters` and `flight::output_parameters columns`, write parameters via `flightworking` table only. 
  - Move serialization/deserialization to `put()`/`get()` time.

### DIFF
--- a/stairway/src/main/java/bio/terra/stairway/FlightDao.java
+++ b/stairway/src/main/java/bio/terra/stairway/FlightDao.java
@@ -712,9 +712,9 @@ class FlightDao {
             flightContext.setFlightStatus(FlightStatus.valueOf(rsflight.getString("status")));
             flightContext.setStepIndex(rsflight.getInt("step_index"));
 
-            // We may have JSON from output_parameters, a set of working parameters from
-            // flightworking table, neither, or both.  Deletgate the decision of which to use to
-            // FlightMap class.
+            // TODO(PF-917): We may have JSON from working_parameters, a set of parameters from
+            // flightworking table, neither, or both.  For now, delegate the decision of which to
+            // use to FlightMap class.  PF-917 will remove column working_parameters.
 
             final String workingMapJson = rsflight.getString("working_parameters");
 
@@ -889,8 +889,9 @@ class FlightDao {
         flightState.setException(
             exceptionSerializer.deserialize(rs.getString("serialized_exception")));
 
-        // We may have JSON from output_parameters, a set of working parameters from flightworking
-        // table, neither, or both.  Deletgate the decision of which to use to FlightMap class.
+        // TODO(PF-917): We may have JSON from output_parameters, a set of parameters from
+        // flightworking table, neither, or both.  For now, delegate the decision of which to use to
+        // FlightMap class.  PF-917 will remove column output_parameters.
 
         String outputParamsJson = rs.getString("output_parameters");
         final List<FlightInput> workingList = retrieveLatestWorkingParameters(connection, flightId);
--- a/stairway/src/main/java/bio/terra/stairway/FlightDao.java
+++ b/stairway/src/main/java/bio/terra/stairway/FlightDao.java
@@ -152,9 +152,9 @@ class FlightDao {
     final String sqlInsertFlightLog =
         "INSERT INTO "
             + FLIGHT_LOG_TABLE
-            + "(id, flightid, log_time, working_parameters, step_index, rerun, direction,"
+            + "(id, flightid, log_time, step_index, rerun, direction,"
             + " succeeded, serialized_exception, status)"
-            + " VALUES (:logId, :flightId, CURRENT_TIMESTAMP, :workingMap, :stepIndex, :rerun, :direction,"
+            + " VALUES (:logId, :flightId, CURRENT_TIMESTAMP, :stepIndex, :rerun, :direction,"
             + " :succeeded, :serializedException, :status)";
 
     String serializedException =
@@ -166,11 +166,6 @@ class FlightDao {
       startTransaction(connection);
       statement.setUuid("logId", logId);
       statement.setString("flightId", flightContext.getFlightId());
-
-      // TODO(PF-703): Column working_parameters is being phased out in favor of table
-      // flightworking.  We continue to write the working map here temporarily for backward
-      // compatibility.
-      statement.setString("workingMap", flightContext.getWorkingMap().toJson());
 
       statement.setInt("stepIndex", flightContext.getStepIndex());
       statement.setBoolean("rerun", flightContext.isRerun());
@@ -397,7 +392,6 @@ class FlightDao {
         "UPDATE "
             + FLIGHT_TABLE
             + " SET completed_time = CURRENT_TIMESTAMP,"
-            + " output_parameters = :outputParameters,"
             + " status = :status,"
             + " serialized_exception = :serializedException,"
             + " stairway_id = NULL"
@@ -411,10 +405,6 @@ class FlightDao {
             new NamedParameterPreparedStatement(connection, sqlUpdateFlight)) {
 
       startTransaction(connection);
-
-      // TODO(PF-703): Column output_parameters is being phased out in favor of table flightworking.
-      //  We continue to write the output map here temporarily for backward compatibility.
-      statement.setString("outputParameters", flightContext.getWorkingMap().toJson());
 
       statement.setString("status", flightContext.getFlightStatus().name());
       statement.setString("serializedException", serializedException);
@@ -722,19 +712,16 @@ class FlightDao {
             flightContext.setFlightStatus(FlightStatus.valueOf(rsflight.getString("status")));
             flightContext.setStepIndex(rsflight.getInt("step_index"));
 
-            // TODO(PF-703): In the current transition away from working_parameters column towards
-            // flightworking table we can either have only JSON or both.  Until we've transitioned,
-            // delegate the decision of which to use to the FlightMap class.
+            // We may have JSON from output_parameters, a set of working parameters from
+            // flightworking table, neither, or both.  Deletgate the decision of which to use to
+            // FlightMap class.
 
             final String workingMapJson = rsflight.getString("working_parameters");
 
             final List<FlightInput> workingList =
                 retrieveWorkingParameters(connection, rsflight.getObject("id", UUID.class));
 
-            // If we built a working map from the data, replace the flight context's default working
-            // map with it.
-            FlightMap.create(workingList, workingMapJson)
-                .ifPresent(workingMap -> flightContext.setWorkingMap(workingMap));
+            flightContext.setWorkingMap(FlightMap.create(workingList, workingMapJson));
           }
         }
       }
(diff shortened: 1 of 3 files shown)


stairway/src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -722,19 +712,16 @@ private void fillFlightContexts(Connection connection, List<FlightContext> fligh
             flightContext.setFlightStatus(FlightStatus.valueOf(rsflight.getString("status")));
             flightContext.setStepIndex(rsflight.getInt("step_index"));
 
-            // TODO(PF-703): In the current transition away from working_parameters column towards
-            // flightworking table we can either have only JSON or both.  Until we've transitioned,
-            // delegate the decision of which to use to the FlightMap class.
+            // We may have JSON from output_parameters, a set of working parameters from
+            // flightworking table, neither, or both.  Deletgate the decision of which to use to
+            // FlightMap class.
```
wchamber: nit: not sure this comment belongs here any more. Seems like this all about FlightMap's promises and the comments belong on that class.
ditto comment on outputParamsJson

stairway/src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -55,130 +46,114 @@ public FlightMap() {
...
     }
     return inputList;
   }
 
   /**
    * Depending on what version of code a Flight log was written with, we may have a JSON Map, a
    * {@code List<FlightInput>}, both, or neither at deserialization time. This method is used to
-   * generate an {@code Optional<FlightMap>} based on what was contained in the database. This will
-   * return an empty {@code Optional<FlightMap>} if json is null.
+   * generate a FlightMap based on what was contained in the database.
```
wchamber, line 58: I'd like an explanation on which version maps to which conditions and especially what the current state is.
Is there a follow up ticket to eventually remove support for the old version, or does it need to be supported indefinitely?

stairway/src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -55,130 +46,114 @@ public FlightMap() {
...
     if (value == null) {
       return null;
     }
 
-    if (type.isInstance(value)) {
-      return type.cast(value);
+    try {
+      return getObjectMapper().readValue(value, type);
+    } catch (JsonProcessingException ex) {
+      throw new ClassCastException(
```
wchamber: this doesn't seem like the right exception anymore. Is there anything that depends on this exception being thrown? (if so, that suggests to me it should be a checked exception). Should we at least propagate the underlying JsonProcessIngException?

wchamber (approved):

stairway/src/main/java/bio/terra/stairway/FlightMap.java
jmczerk: I updated the PR description to link out to [the document](https://docs.google.com/document/d/1b1VckTBb_V1pCMhoN2GgHnGDR2f8MV2HuwIgDqFc_9M/edit) describing the phases of this change.  

Current services are all in Phase 1 (last to go was TDR back in mid-May), where we are writing both the working/output parameter fields as JSON maps and writing to the new `flightworking` table.  We are also doing informational checks to make sure that the working/output maps and `flightworking` contents match up (where `flightworking` rows are present).

These checks are informational only and log ERRORs (but do not result in failures).  There was actually a bug in the verification logic that showed some false positives (PF-709) which @ddietterich observed in WSM testing, which I think gives us some confidence that we are not seeing mismatches here.

I created a ticket [PF-917] for removal of these columns.  This was initially slated as "not POR" for this process due to the number of old flights hanging around (esp in TDR), but there may be an opportunity to do this sooner now that we have implemented grooming of old flights.

stairway/src/main/java/bio/terra/stairway/FlightMap.java
jmczerk: I searched and don't see anyone catching `ClassCastException` in this context.  So perhaps throw (unchecked) `JsonConversionException` makes more sense here? 

@ddietterich: interested in your input given your plans to rationalize exceptions in your upcoming reorg work.

jmczerk: I searched and don't see anyone catching ClassCastException in this context. So perhaps throw (unchecked) JsonConversionException makes more sense here?

@ddietterich: interested in your input given your plans to rationalize exceptions in your upcoming reorg work.

stairway/src/main/java/bio/terra/stairway/FlightDao.java
jmczerk: I feel like having this in both places doesn't hurt, as it explains why both columns exist in the data model without having to cross-reference `FlightMap`.  That said, if you feel like the docs in `FlightMap` are insufficient I can try to improve them (they look OK to me, but I have the full context in my head).

ddietterich: I think it is worth noting here. It would be worth mentioning PF-917 that would remove the output_parameters and working_parameters fields.

stairway/src/main/java/bio/terra/stairway/FlightMap.java
ddietterich: IIUC, if all Stairway clients have moved to Phase 1, then there is no timing issue with installing Phase 2.

When we do PF-917, then clients need to ensure that there are no pre-phase 2 flights in their databases. We can add a `stairctl` command to do that check.

stairway/src/main/java/bio/terra/stairway/FlightMap.java
ddietterich: I agree that JsonConversionException makes more sense.

ddietterich (approved):
LGTM

stairway/src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -55,130 +46,123 @@ public FlightMap() {
...
+   *
+   * <p>Flights written with:
+   *
+   * <p>* Version < 0.0.50: Only json is valid, inputList is always empty.
+   *
+   * <p>* Version 0.0.50 - 0.0.60: json and inputList both valid and must be consistent with one
+   * another.
+   *
+   * <p>* Version >= 0.0.61: json always null, inputList is always valid (possibly empty).
    *
```
jmczerk, line 68: @wchamber : Pls let me know if this is helpful!

stairway/src/main/java/bio/terra/stairway/FlightDao.java
jmczerk: Docs now reference PF-917.

stairway/src/main/java/bio/terra/stairway/FlightMap.java
jmczerk: Phase 2 is not backward compatible, so there's no rolling back from this (which is why we were conservative and did the schema update separately then let this run a while with the ERROR logging in place).

Am I correct in my assumption that absent a rollback, we do not have any scenarios where we will try to read a flight written with Phase 2 code using Phase 1 code?

ddietterich: That is my understanding also.

stairway/src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -55,130 +46,123 @@ public FlightMap() {
...
       return null;
     }
 
-    if (type.isInstance(value)) {
-      return type.cast(value);
+    try {
+      return getObjectMapper().readValue(value, type);
+    } catch (JsonProcessingException ex) {
+      throw new JsonConversionException(
+          "Found value '" + value + "' is not an instance of type " + type.getName());
```
wchamber: I don't think it's right to say that the value isn't an instance of the type, just that there was some problem processing the json. I also think it'd be useful to include the thrown `ex` as the cause for this exception for debugging. instead of throwing it away.

wchamber (approved):

stairway/src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -55,130 +46,123 @@ public FlightMap() {
...
    *
    * @param <T> - type of class to expect in the hash map
    * @param key - key to lookup in the hash map
    * @param type - class requested
    * @return null if not found
-   * @throws ClassCastException if found, not castable to the requested type
+   * @throws ClassCastException if found, not deserializable to the requested type
    */
+  @Nullable
   public <T> T get(String key, Class<T> type) {
```
pshapiro4broad, line 108: Now that you're calling ObjectMapper on a per-field basis, you could add an overloaded `get()` method that accepts `TypeReference<T>` instead of `Class<T>`. That would make it possible to deserialize generics with type safety, and users can rely on type inference instead of specifying the type as a parameter.

Another nice feature would be to return `Optional<T>` instead of `T` which would make the method return policy clearer. But that would require making a new API and deprecating this one, so it's probably not worth it now.

stairway/src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -55,130 +46,123 @@ public FlightMap() {
...
+   * generate a FlightMap based on what was contained in the database.
+   *
+   * <p>Flights written with:
+   *
+   * <p>* Version < 0.0.50: Only json is valid, inputList is always empty.
+   *
+   * <p>* Version 0.0.50 - 0.0.60: json and inputList both valid and must be consistent with one
+   * another.
+   *
+   * <p>* Version >= 0.0.61: json always null, inputList is always valid (possibly empty).
```
pshapiro4broad: Very minor point; it looks like you're trying to create an unordered list here, if so, it would be be better to use `ul`, `li`.
```suggestion
   * <p>Flights written with:
   *
   * <ul><li>Version < 0.0.50: Only json is valid, inputList is always empty.
   *
   * <li>Version 0.0.50 - 0.0.60: json and inputList both valid and must be consistent with one
   * another.
   *
   * <li>Version >= 0.0.61: json always null, inputList is always valid (possibly empty).
   * </ul>
```

stairway/src/main/java/bio/terra/stairway/FlightMap.java
jmczerk: Thanks, cleaned this up, as well as the missed comments.

stairway/src/main/java/bio/terra/stairway/FlightMap.java
jmczerk: Thanks, this was helpful!

stairway/src/main/java/bio/terra/stairway/FlightMap.java
jmczerk: Thanks for this feedback...  I'd like to explore this a bit more to get a better feel for the benefits we can get from this, would it be OK if we continued this thread in a follow-on PR so I can start propagating the in-database changes downstream?

pshapiro4broad: > would it be OK if we continued this thread in a follow-on PR so I can start propagating the in-database changes downstream?

That's fine with me.

> I'd like to explore this a bit more to get a better feel for the benefits we can get from this

There's been a bunch of articles written about how this API is better than `Class<T>`, here's one. https://northcoder.com/post/jackson-object-mapper-which-way-is/

(Looking at this link again, I'm not sure it captures the use case where is where the type argument to `TypeReference` is inferred from the return type of the `readValue()` method call. e.g.
```
        List<String> list = objectMapper.readValue("[\"hello\"]", new TypeReference<>() { });
```
)

PR #76, opened 2021-05-19
[PF-785] Fix scoping on subquery when getting latest working parameters.

jmczerk:
This query fails:
```
        "SELECT key, value FROM "
            + FLIGHT_WORKING_TABLE
            + " WHERE flightlog_id="
            + "(SELECT id FROM flightlog WHERE log_time="
            + "   (SELECT MAX(log_time) FROM flightlog WHERE flightid = :flightId))";
```
Because `flightlog::id` can be NULL, so multiple rows can be returned from:

`SELECT id FROM flightlog WHERE log_time=`

Fix is to also select on `flightid` here.

### DIFF
--- a/src/main/java/bio/terra/stairway/FlightDao.java
+++ b/src/main/java/bio/terra/stairway/FlightDao.java
@@ -926,8 +926,8 @@ class FlightDao {
         "SELECT key, value FROM "
             + FLIGHT_WORKING_TABLE
             + " WHERE flightlog_id="
-            + "(SELECT id FROM flightlog WHERE log_time="
-            + "   (SELECT MAX(log_time) FROM flightlog WHERE flightid = :flightId))";
+            + "(SELECT id FROM flightlog WHERE (flightid = :flightId) AND log_time="
+            + "   (SELECT MAX(log_time) FROM flightlog WHERE flightid = :flightId2))";
 
     List<FlightInput> inputList = new ArrayList<>();
 
@@ -935,6 +935,7 @@ class FlightDao {
         new NamedParameterPreparedStatement(connection, sqlSelectInput)) {
 
       statement.setString("flightId", flightId);
+      statement.setString("flightId2", flightId);
       try (ResultSet rs = statement.getPreparedStatement().executeQuery()) {
         while (rs.next()) {
           FlightInput input = new FlightInput(rs.getString("key"), rs.getString("value"));


src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -926,15 +926,16 @@ private void storeWorkingParameters(
         "SELECT key, value FROM "
             + FLIGHT_WORKING_TABLE
             + " WHERE flightlog_id="
-            + "(SELECT id FROM flightlog WHERE log_time="
-            + "   (SELECT MAX(log_time) FROM flightlog WHERE flightid = :flightId))";
+            + "(SELECT id FROM flightlog WHERE (flightid = :flightId) AND log_time="
+            + "   (SELECT MAX(log_time) FROM flightlog WHERE flightid = :flightId2))";
```
nmalfroy, line 930: Why does this need to be a new substitution variable?   : flightId and :flightId2 both get set to `flightId`

wchamber (approved):
My reading is that flightid is only null in older versions of the database, but if that's not the case (i.e. we can still have that happen in this version of the code) might be good to add a test.

src/main/java/bio/terra/stairway/FlightDao.java
jmczerk: I thought so too, but our [`NamedParameterPreparedStatement`](https://github.com/DataBiosphere/stairway/blob/develop/src/main/java/bio/terra/stairway/NamedParameterPreparedStatement.java) class will only substitute the first instance seen.  Fortunately unit tests caught it.

Probably worth considering enhancing `NamedParameterPreparedStatement` as a follow on, I can create a Jira ticket.  In the meantime this is a faster path to a fix and [there's precedent](https://github.com/DataBiosphere/stairway/blob/develop/src/main/java/bio/terra/stairway/FlightDao.java#L636).

jmczerk:
> My reading is that flightid is only null in older versions of the database, but if that's not the case (i.e. we can still have that happen in this version of the code) might be good to add a test.

Assuming you mean `flightlog::id`...  Yeah, I couldn't make this non-nullable AND unique, because postgres can't generate default GUID's without a plugin, so existing flight logs will have null ID's (after upgrade -- the column didn't exist before).  There's no way to get a NULL flight log ID using Stairway proper code; I can write a test that hacks the DB directly, but that will slow delivery of this fix (there's not a lot of precedent and thus plumbing for this).  I'll probably add a ticket for follow-on work.

src/main/java/bio/terra/stairway/FlightDao.java
ddietterich: `NamedParameterPreparedStatement` is not a sophisticated implementation. I made it that way so that I didn't have to touch the values or deal with variable types.

ddietterich (approved):
LGTM

ddietterich:
Another follow-on work thing might be to make sure we never write two log records with the same timestamp. It is a small PITA, but not really too hard to do.

PR #77, opened 2021-05-20
[PF-785] Regression test for duplicate timestamp fix.

jmczerk:
Draft PR to gauge interest in whether this should be checked in.

Basically without the fix for PF-785, I was able to repro the bug about 20% of the time when running this test on my Mac, as it does induce `flightlog` rows with duplicate timestamps across different flights.  Manually inspecting the database after running this test shows that the fix is effective when duped timestamps are present.

### DIFF
--- a/src/test/java/bio/terra/stairway/ScenarioTest.java
+++ b/src/test/java/bio/terra/stairway/ScenarioTest.java
@@ -22,8 +22,11 @@ import bio.terra.stairway.flights.TestFlightUndo;
 import bio.terra.stairway.flights.TestFlightWait;
 import java.io.File;
 import java.io.PrintWriter;
+import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.List;
 import java.util.UUID;
+import java.util.concurrent.Phaser;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


src/test/java/bio/terra/stairway/ScenarioTest.java
```diff
@@ -715,4 +718,109 @@ private String makeExistingFile() throws Exception {
...
+        FlightMap inputParameters = new FlightMap();
+        inputParameters.put("filename", filename);
+        inputParameters.put("text", "testing 1 2 3");
+
+        logger.debug(
+            "Waiting to launch TestFlight '{}' with filename '{}'", getFlightId(), filename);
+
+        // Phaser is used here as a simple barrier to try to induce duplicate timestamps.  All
+        // threads wait on the barrier until the last one reaches it.
+        phaser.arriveAndAwaitAdvance();
```
ddietterich, line 760: How did I not know about Phaser?! That is really cool.
I should just know that any possible problem I have in Java has already been solved. :)

ddietterich (approved):
LGTM, thanks.

marikomedlock (approved):
LGTM. I'd also not heard of Phaser before, so thanks for the introduction!

PR #78, opened 2021-05-21
Add workflow_dispatch trigger to develop version bump.

jmczerk:
I would have liked to make the bump+publish trigger a little more full featured[ like Workspace Manager's](https://github.com/DataBiosphere/terra-workspace-manager/blob/dev/.github/workflows/tag-publish.yml#L30), but this is really a hedge against recent flakiness in GitHub actions.  PF-789 created for future improvements.

### DIFF
--- a/.github/workflows/bumpDevelop.yaml
+++ b/.github/workflows/bumpDevelop.yaml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - develop
+  workflow_dispatch: {}
 jobs:
   update_image:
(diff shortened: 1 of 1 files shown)


ddietterich (approved):
Thanks for wrangling such a massive change! :)

yonghaoy (approved):

jmczerk:
Still not authorized to merge 😞 

Screen Shot 2021-05-21 at 3 10 30 PM

PR #79, opened 2021-06-08
PF-312 grooming, visibility and control interface

ddietterich:
This PR adds grooming to Stairway

Currently, Stairway keeps all flight data forever. The grooming feature provides client control over how long completed flight data is retained. Two parameters are used to control grooming:
- `completedFlightRetention` specifies how long Stairway keeps completed flight data
- `retentionCheckInterval` specifies how often Stairway checks and deletes old flights
If not specified, the current behavior is maintained.

With the availability of grooming, the old parameter `keepFlightLog` is deprecated and its setting is ignored.


src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -52,32 +54,39 @@
...
-      boolean keepFlightLog) {
+      Duration completedFlightAvailable) {
     this.dataSource = dataSource;
     this.stairwayInstanceDao = stairwayInstanceDao;
     this.exceptionSerializer = exceptionSerializer;
     this.hookWrapper = hookWrapper;
-    this.keepFlightLog = keepFlightLog;
+
+    int visibility = 0;
+    if (completedFlightAvailable != null) {
```
jmczerk, line 82: Is the intention that a null `completedFlightAvailable` implies "keep forever"?  If so it might be good to make this more explicit via doc/`@nullable`/`Optional<>` (no strong opinion on how).

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -483,6 +492,70 @@ private void deleteInner(String flightId) throws SQLException {
     }
   }
 
+  /**
+   * Remove completed flights from the database that are older than a specific time
+   *
+   * @param deleteOlderThan time before which flights can be removed
+   * @throws DatabaseOperationException on any database error
+   * @throws InterruptedException thread shutdown
```
jmczerk, line 487: Nit: `@returns` -> `@return`

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -52,32 +54,39 @@
...
       HookWrapper hookWrapper,
-      boolean keepFlightLog) {
+      Duration completedFlightAvailable) {
     this.dataSource = dataSource;
     this.stairwayInstanceDao = stairwayInstanceDao;
     this.exceptionSerializer = exceptionSerializer;
     this.hookWrapper = hookWrapper;
-    this.keepFlightLog = keepFlightLog;
+
+    int visibility = 0;
```
jmczerk: Is the intention that a null `completedFlightAvailable` implies "all flights visible"? If so it might be good to make this more explicit via doc/`@Nullable`/`Optional<>` (no strong opinion on which).

src/main/java/bio/terra/stairway/CompletedFlightCleaner.java
```diff
@@ -0,0 +1,35 @@
...
+
+  @Override
+  public void run() {
+    Instant deleteOlderThan = Instant.now().minus(retention);
+    logger.info("Removing flights completed before {}", deleteOlderThan);
+    try {
+      int count = flightDao.deleteCompletedFlights(deleteOlderThan);
+      logger.info("Cleaned up {} completed flights", count);
+    } catch (DatabaseOperationException | InterruptedException ex) {
+      // Simply return
```
wchamber: probably worth logging the exception

src/main/java/bio/terra/stairway/Control.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public class Control {
+  private final FlightDao flightDao;
+
+  public Control(FlightDao flightDao) {
+    this.flightDao = flightDao;
+  }
+
+  /**
+   * Lookup a flight without the completed time constraint
```
wchamber: Probably worth stating what the completed time constraint is and why it's useful to avoid it.

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -682,32 +755,54 @@ private void fillFlightContexts(Connection connection, List<FlightContext> fligh
...
             + FLIGHT_TABLE
             + " WHERE flightid = :flightId";
 
+    boolean enforceVisibility = enforceVisibilityLimit && (flightVisibilitySeconds > 0);
+
+    String sqlOneFlight = sqlOneFlightBase;
+    if (enforceVisibility) {
+      sqlOneFlight =
+          sqlOneFlightBase
+              + " AND (completed_time IS NULL OR completed_time + CAST(:visibility AS INTERVAL) > CURRENT_TIMESTAMP)";
```
wchamber: nit: Maybe inline the ':visibility seconds' or move the whole interval cast to the oneFlightStatement.setString? As-is the split between here and string.format was a little surprising to me.

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -52,32 +54,39 @@
...
     this.exceptionSerializer = exceptionSerializer;
     this.hookWrapper = hookWrapper;
-    this.keepFlightLog = keepFlightLog;
+
+    int visibility = 0;
+    if (completedFlightAvailable != null) {
+      try {
+        visibility = Math.toIntExact(completedFlightAvailable.getSeconds());
+      } catch (ArithmeticException e) {
+        // Treat huge duration as equivalent to forever
```
wchamber: worth a log warning when this happens?

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -682,32 +755,54 @@ private void fillFlightContexts(Connection connection, List<FlightContext> fligh
   }
 
   /**
-   * Return flight state for a single flight
+   * Return flight state for a single flight. The status of completed flight is limited by
+   * completeFlightAvailable property.
```
wchamber, line 746: I don't see the `completeFlightAvailable` property mentioned anywhere else.

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -682,32 +755,54 @@ private void fillFlightContexts(Connection connection, List<FlightContext> fligh
...
+   * @throws DatabaseOperationException - unexpected error in the operation
+   * @throws InterruptedException - interrupt
    */
   FlightState getFlightState(String flightId)
       throws DatabaseOperationException, InterruptedException {
-    return DbRetry.retry("flight.getFlightState", () -> getFlightStateInner(flightId));
+    return DbRetry.retry("flight.getFlightState", () -> getFlightStateInner(flightId, true));
+  }
+
+  FlightState controlGetFlightState(String flightId)
```
wchamber: Worth a comment on how this is different from `getFlightState`. `control` doesn't really explain much to me on its own.

src/main/java/bio/terra/stairway/Stairway.java
```diff
@@ -233,6 +240,45 @@ public Builder workQueueSubscriptionId(String workQueueSubscriptionId) {
       return this;
     }
 
+    /**
+     * Control the frequency of clean up passes over the retained flights. Defaults to never
+     * checking.
```
wchamber: Defaulting to checking once an hour or day seems better. This can still not do anything if no retention is set.

wchamber (commented):
I'm not totally sold on 'visibility' yet. When is it helping me? When I don't want to return the FlightState, but the FlightState can still be deserialized and should not yet be deleted? I'm not really following the use case that makes it worth the complexity of having invisible flights.

If we're making a breaking change, an interval isn't going to let us mark a cut in time. If the change isn't so breaking that we need a sharp divide, why not just use the grooming flight deletion?

If we want a complete log of everything ever, dumping regularly to a separate table/database at regular intervals seems simpler to reason about than invisibility on the active table.

ddietterich:
> I'm not totally sold on 'visibility' yet. When is it helping me? When I don't want to return the FlightState, but the FlightState can still be deserialized and should not yet be deleted? I'm not really following the use case that makes it worth the complexity of having invisible flights.
> 
> If we're making a breaking change, an interval isn't going to let us mark a cut in time. If the change isn't so breaking that we need a sharp divide, why not just use the grooming flight deletion?
> 
> If we want a complete log of everything ever, dumping regularly to a separate table/database at regular intervals seems simpler to reason about than invisibility on the active table.

I'm not sure about it either and would be OK dropping it. @jmczerk please give your thoughts as well. I've been thinking about it two ways.

First, from the point of view of a service exposing an async API, how long do you want to be on the hook for returning results? If another service asks for results that are a month old, is that reasonable or a sign of some other issue? If we want services to be able to bound that, then having Stairway support it rather than each service would be reasonable. Is there an analogy with something Google does that we could take inspiration from?

Second, when we finish the serdes changes, Stairway will not be deserializing as part of building up the FlightState object. So it means we can have things in the database that are no longer deserializable. When we do a migration from A to (A and B) to B, we need to know when we can deploy B. It would nice not to wait 60 days to do that. You can compute when it is safe to deploy B by finding the last completed A flight and then waiting the interval.

wchamber:
For my 2c:
1: I think the flight grooming handles how long I want to be on the hook for returning results. ~1 week for debugging, but realistically order hours for down stream clients. Job results aren't long term storage.
2: I think that's on the operating service. If you want to make a breaking change, see if there are any outstanding flights from before you started your migration (which is probably some kind of issue). And wait at least until you're not on the hook for returning results (which you have to do to satisfy your contracts to clients regardless if the results are removed by visibility or grooming). So I would push this into operational toil.

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -483,6 +492,70 @@ private void deleteInner(String flightId) throws SQLException {
...
+   * @throws InterruptedException thread shutdown
+   * @returns count of deleted lights
+   */
+  int deleteCompletedFlights(Instant deleteOlderThan)
+      throws DatabaseOperationException, InterruptedException {
+    return DbRetry.retry(
+        "flight.deleteCompletedFlights", () -> deleteCompletedFlightsInner(deleteOlderThan));
+  }
+
+  private int deleteCompletedFlightsInner(Instant deleteOlderThan) throws SQLException {
```
jmczerk, line 496: Will this only be run by grooming, or do you expect it to also be called by `stairctl`?  Also any concerns about a flood of deletions on existing DB's when grooming gets turned on that might raise a desire to limit deletions per cycle?

jmczerk:
> First, from the point of view of a service exposing an async API, how long do you want to be on the hook for returning results? If another service asks for results that are a month old, is that reasonable or a sign of some other issue? If we want services to be able to bound that, then having Stairway support it rather than each service would be reasonable. Is there an analogy with something Google does that we could take inspiration from?

I agree with Will that it sees very reasonable to limit how long a flight result can be expected to be returned, as long as it’s well understood by the client.   I strongly agree that given Stairway’s primary purpose (transactional safety) and usage on important code paths, I would prefer to not see the stairway DB used for long term storage or auditing.

> Second, when we finish the serdes changes, Stairway will not be deserializing as part of building up the FlightState object. So it means we can have things in the database that are no longer deserializable. When we do a migration from A to (A and B) to B, we need to know when we can deploy B. It would nice not to wait 60 days to do that. You can compute when it is safe to deploy B by finding the last completed A flight and then waiting the interval.

I see potential debugging benefit of allowing things to stick around beyond when they can be expected to be read (under normal operations), and I don't feel like the added complexity is onerous.  The alternative suggestion of allowing things to fall off into an archive table/database at grooming time also sounds like another way to provide a similar debugging capability (and maybe even make allowances for long-term storage/auditing), but I suspect it would be of at least a similar level of complexity.

ddietterich:
Thank you both for your considered thoughts on this. I'll remove the visibility part and just include the grooming. If we get a clearer indication that holding data for longer is useful for debug or audit, we can solve that problem separately.

src/main/java/bio/terra/stairway/Stairway.java
ddietterich: Fair enough.

src/main/java/bio/terra/stairway/FlightDao.java
ddietterich: The implementation only runs this as a periodic background task so far. `stairctl` is just a prototype right now. I was waiting until serdes moves to FlightMap access time to make it real, because I didn't want to support a parallel DAO implementation.

Anyway, I wasn't being concerned about a big delete transaction, but maybe I should be?!

Since there is no index on the completed time, the delete will cause a full table scan regardless of the number of rows involved. That will cause it to conflict with the short write transaction typical of the Stairway workload. The delete will have some index updates to do for the PK. It could be a transaction that would run long enough that it would cause a conflicting transaction to fail despite retrying.

I'm not confident that deleting fewer rows at a time would help enough, though. The full table scan would still be there. IIRC inserts are used to express deletions, so that is sequentialized writing to the tail of the table. I have no clue how the index updating is done. That would be the most obvious part of the plan that would scale directly with the number of rows.

An alternate approach to rolling this out would be to build `stairctl` with a "groom since <timestamp>" command. That would allow a human to run the deletes on off hours and/or do them in smaller chunks. Then the automatic grooming could be enabled. That would be fun, but is a good bit of effort for a one time event.

ddietterich: I think I would like to merge this doing the delete in the straightforward way and get started on the common lib work.

src/main/java/bio/terra/stairway/FlightDao.java
```diff
@@ -784,11 +848,22 @@ private FlightState getFlightStateInner(String flightId)
...
+      throws DatabaseOperationException, InterruptedException {
+
+    // Make an empty filter if one is not provided
+    FlightFilter filter = (inFilter != null) ? inFilter : new FlightFilter();
+
+    return DbRetry.retry("flight.getFlights", () -> getFlightsInner(offset, limit, filter, false));
+  }
+
+  private List<FlightState> getFlightsInner(
+      int offset, int limit, FlightFilter filter, boolean enforceVisibilityLimit)
```
wchamber: I think this is an incomplete revert?

src/test/java/bio/terra/stairway/FlightCleanerTest.java
```diff
@@ -0,0 +1,69 @@
...
+  @Test
+  public void successTest() throws Exception {
+    final String stairwayName = "flightCleanerTest";
+
+    // Start with a clean and shiny database environment.
+    Stairway stairway =
+        Stairway.newBuilder()
+            .stairwayName(stairwayName)
+            .completedFlightRetention(Duration.ofMinutes(1))
+            .retentionCheckInterval(Duration.ofSeconds(15))
```
wchamber: could we make these even shorter to let the test complete faster?

wchamber (approved):

src/main/java/bio/terra/stairway/FlightDao.java
ddietterich: Yup, thank you

src/test/java/bio/terra/stairway/FlightCleanerTest.java
ddietterich: Good idea. I separated the numbers out into constants and shrunk the time a bit.

src/main/java/bio/terra/stairway/FlightDao.java
jmczerk: Does this still need to be removed?

src/main/java/bio/terra/stairway/FlightFilter.java
```diff
@@ -241,7 +241,8 @@ private void makeSqlForm3(StringBuilder sb) {
 
   /**
    * Generate the filter for the flight table. The right hand side of any comparison op is generated
-   * as a named parameter.
+   * as a named parameter. This is the method that injects or does not inject the flight visibility
```
jmczerk: Stray ref to flight visibility?

jmczerk (approved):
Couple missed cleanups from the revert, but otherwise looks good.

PR #80, opened 2021-06-10
[PF-827] Fix to validation of flightworking vs flight map

ddietterich:
The validator was not distinguishing between "key not found" and "key found and value is null".

I verified the fix by changing WSM to build against mavenLocal and running the previously-failing test against the change.
The spurious error was not reported.


jmczerk (approved):
Thanks for the quick turnaround on this!

wchamber (approved):

PR #81, opened 2021-06-17
[PF-501] stairctl

ddietterich:
This PR implements the `stairctl` and documents it in STAIRCTL.md

I reorganized the code into composite build structure, so what used to be the only project is now in the `library` directory and there is a new project for `stairctl`.  I don't think `stairctl` needs a close review, but I want to be sure this won't break anything in Stairway.

Stairway gets a new class `Control` that can be accessed from the `Stairway` object via `getControl()`. It is implements database queries that map pretty close to 1:1 to commands in `stairctl`.


library/src/main/java/bio/terra/stairway/Control.java
```diff
@@ -0,0 +1,583 @@
...
+
+  // -- Flight methods --
+
+  private static final String flightSelectFrom =
+      " flightid, submit_time, class_name, completed_time, status,"
+          + "serialized_exception, stairway_id FROM flight ";
+
+  private static final String flightOrderPage =
+      " ORDER BY submit_time DESC OFFSET :offset LIMIT :limit";
+
```
jmczerk, line 54: Maybe a `static final String` for the `SELECT COUNT(*) AS total FROM flight` for use in `countQuery()` callers too?

library/src/main/java/bio/terra/stairway/Control.java
```diff
@@ -0,0 +1,583 @@
...
+    fakeFlightContext.setFlightStatus(FlightStatus.READY);
+    fakeFlightContext.setFlightId(flightId);
+    flightDao.exit(fakeFlightContext);
+
+    return getFlight(flightId);
+  }
+
+  public Control.Flight forceFatal(String flightId) throws SQLException {
+    // We do not assume anything about the current state. We just force to a FATAL
+    // disowned state so no attempt will be made to recover.
```
jmczerk, line 171: re: the `TODO`, do you mean so we can tell the difference between a forced FATAL and a natural FATAL?

stairctl/src/main/java/bio/terra/stairctl/commands/ConnectCommands.java
```diff
@@ -0,0 +1,65 @@
+package bio.terra.stairctl.commands;
+
+import bio.terra.stairctl.ConnectParams;
+import bio.terra.stairctl.StairwayService;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.shell.standard.ShellComponent;
+import org.springframework.shell.standard.ShellMethod;
+import org.springframework.shell.standard.ShellOption;
+
```
jmczerk, line 9: Spring Shell is pretty cool, I didn't know this existed...  So the only state we track during the session currently connection state, but if we needed to track more session state would it go in `StairwayService`?

stairctl/src/main/java/bio/terra/stairctl/commands/Output.java
```diff
@@ -0,0 +1,222 @@
...
+import java.util.Comparator;
+import java.util.List;
+import java.util.Optional;
+import javax.annotation.Nullable;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
```
jmczerk, line 20: I'm terrible at envisioning format outputs in my head, if you don't mind posting a couple examples of the  output of chattier commands, that would be helpful.

stairctl/src/main/java/bio/terra/stairctl/commands/FlightCommands.java
```diff
@@ -0,0 +1,179 @@
...
+  @ShellMethodAvailability({
+    "count flights",
+    "count owned",
+    "force fatal",
+    "force ready",
+    "get flight",
+    "list dismal",
+    "list flights",
+    "list owned"
+  })
```
jmczerk, line 44: If we expected more `*Commands` that needed this OR it got much more complex, I might think about using a mixin to share this code, but for now it's probably OK to have two copies.

jmczerk (approved):
This looks really useful and nicely done. 

Also quickly clicking through 100+ `File renamed without changes.` files and clicking "Viewed" is oddly satisfying, kind of like popping bubble wrap...

library/src/main/java/bio/terra/stairway/Control.java
ddietterich: Yes. I am not sure what the right thing is, but it seems like we should be keeping an audit trail somewhere of the changes made. Especially in a production situation. I made a ticket for this and referenced it here in the TODO.

stairctl/src/main/java/bio/terra/stairctl/commands/ConnectCommands.java
ddietterich: Probably, although we could add other `@Components` if we had other kinds of state.

stairctl/src/main/java/bio/terra/stairctl/commands/Output.java
ddietterich: Would it be OK if I put that in the STAIRCTL.md? Or would you like it in-line as comments?

jmczerk: STAIRCTL.md would be perfect

stairctl/src/main/java/bio/terra/stairctl/commands/FlightCommands.java
ddietterich: You'll have to explain what you mean by `mixin` sometime.

jmczerk: In this case it'd just mean an interface with a default implementation of `availabilityCheck()`; `*Command` would implement the interface and take the default impl, as opposed to inheriting from some an abstract with `availabilityCheck()`  (hopefully @wchamber will correct me if I'm abusing the term).

library/src/main/java/bio/terra/stairway/Control.java
```diff
@@ -0,0 +1,583 @@
...
+      }
+      if (direction != logEntry.direction) {
+        return false;
+      }
+      return id != null ? id.equals(logEntry.id) : logEntry.id == null;
+    }
+
+    @Override
+    public int hashCode() {
+      int result = flightId != null ? flightId.hashCode() : 0;
```
pshapiro4broad: Is there a reason why you don't use `Objects.hash`? I think it would make this method much simpler, and provide equivalent functionality.

library/src/main/java/bio/terra/stairway/Control.java
```diff
@@ -0,0 +1,583 @@
...
+
+      LogEntry logEntry = (LogEntry) o;
+
+      if (stepIndex != logEntry.stepIndex) {
+        return false;
+      }
+      if (rerun != logEntry.rerun) {
+        return false;
+      }
+      if (flightId != null ? !flightId.equals(logEntry.flightId) : logEntry.flightId != null) {
```
pshapiro4broad: This can be done using `Objects.equals`, which handles the null check too.

library/src/main/java/bio/terra/stairway/Control.java
```diff
@@ -0,0 +1,583 @@
...
+      StairwayInstanceDao stairwayInstanceDao) {
+    this.stairway = stairway;
+    this.dataSource = dataSource;
+    this.flightDao = flightDao;
+    this.stairwayInstanceDao = stairwayInstanceDao;
+  }
+
+  // -- Flight methods --
+
+  private static final String flightSelectFrom =
```
pshapiro4broad: These constant names don't follow the google style guidelines. Is there a reason to not follow the guidelines here?

https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

library/src/main/java/bio/terra/stairway/Control.java
```diff
@@ -0,0 +1,583 @@
...
+      int flightCount = countQuery(statement);
+      DbUtils.commitTransaction(connection);
+      return flightCount;
+    }
+  }
+
+  public int countOwned() throws SQLException {
+    final String sql = "SELECT COUNT(*) AS total FROM flight WHERE stairway_id IS NOT NULL";
+    try (Connection connection = dataSource.getConnection();
+        NamedParameterPreparedStatement statement =
```
pshapiro4broad: This is a case where I think `var` can be used to make the code shorter (and therefore easier to read) without sacrificing clarity. 
```
var statement =
```

pshapiro4broad (commented):
I think this is a great idea and I'm looking forward to trying it out with TDR. Did you consider adding an option to make the output easily parseable (e.g., JSON) so its output could be consumed by other programs?

library/src/main/java/bio/terra/stairway/Control.java
ddietterich: I took the IntelliJ default. I regenerated with the Objects style. Very nice. Thanks for suggesting it!

ddietterich:
> I think this is a great idea and I'm looking forward to trying it out with TDR. Did you consider adding an option to make the output easily parseable (e.g., JSON) so its output could be consumed by other programs?

Thanks so much for taking a look at the PR! And as usual, you teach me something new about Java.

The Spring Shell puts you at its own interactive prompt. It is like running `psql`. So writing JSON output wouldn't be easily consumable. If there is a need for that, it could be done. We would need to add some command line processing (like `-c` in `psql`) that would connect, run a command, and then exit. I have read stackoverflow conversations about how to do that - it is possible.

STAIRCTL.md
```diff
@@ -0,0 +1,151 @@
...
+`-o` or `--offset` for the start of the page and `-l` or `--limit` for the number of flights
+to list.
+
+* `list flights` command provides an optional `-s` or `--status` option to filter for flights in
+specific states. For example, `list flights --status QUEUED`.
+
+* `list dismal` command is a shortcut for `list flights --status FATAL`; providing a list
+of flights that require manual intervention for recovery.
+
+* `list owned` command shows flights that are currently owned by a Stairway instance. Those might
```
wchamber: not necessarily this PR, but is there a way to filter to flights that are not owned? or flights owned by a particular instance? Probably not as important to start with.

STAIRCTL.md
```diff
@@ -0,0 +1,151 @@
...
+By default, the `get flight` command provides a summary of the flight. It takes a flight id.
+You can get more data by setting switches:
+* `--input` will display the input paramters
+* `--log` will display a tabular summary of the log
+* `--logmap` will display details of the log including the working map at each point
+For example, `get flight --input --logmap --flight-id a63fc80f-f8a7-4cbc-be58-6cebc98ac21f`
+will provide the maximum, noisiest retrieval.
+
+## Flight Use Cases
+This section provides a few capabilities of StairCtl.
```
wchamber, line 274: thanks, this is really helpful.

library/src/main/java/bio/terra/stairway/Control.java
```diff
@@ -0,0 +1,583 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.stream.Collectors;
+import javax.sql.DataSource;
+
+@SuppressFBWarnings(
+    value = "RCN_REDUNDANT_NULLCHECK_WOULD_HAVE_BEEN_A_NPE",
+    justification = "Spotbugs doesn't understand resource try construct")
+public class Control {
```
wchamber, line 29: nit: worth a comment on what this class is for and why we want it separated from FlightDao.

library/src/main/java/bio/terra/stairway/Control.java
```diff
@@ -0,0 +1,583 @@
...
+    int flightCount = 0;
+    try (ResultSet rs = statement.getPreparedStatement().executeQuery()) {
+      while (rs.next()) {
+        flightCount = rs.getInt("total");
+      }
+    }
+    return flightCount;
+  }
+
+  public int countFlights(String status) throws SQLException {
```
wchamber: nit: any particular reason to take a String instead of a FlightStatus and leave parsing FlightStatus to clients? Ditto others. if it's particularly helpful, we could provide a static conversion or throw utility and deduplicate some code.

library/src/main/java/bio/terra/stairway/Control.java
```diff
@@ -0,0 +1,583 @@
...
+
+    public Flight stairwayId(String stairwayId) {
+      this.stairwayId = stairwayId;
+      return this;
+    }
+  }
+
+  // Holds flight map data as a pair of strings; either inputs or working map
+  // Comparable by key for an alphabetical display
+  public static class KeyValue implements Comparable<KeyValue> {
```
wchamber: optional nit: consider `FlightMapEntry`? Little more specific about what this is for.

stairctl/build.gradle
```diff
@@ -0,0 +1,106 @@
...
+        junit = "5.7.1"
+        logback = "1.2.3"
+        sl4j = "1.7.30"
+        springboot = "2.4.5"
+        springshell = "2.0.1.RELEASE"
+    }
+
+    //
+    implementation project(":library")
+    //implementation files("/Users/ddietterich/codestair/PF-501-stairctl/library/build/libs/library-0.0.56-SNAPSHOT.jar")
```
wchamber: remove me before merge?

stairctl/build.gradle
```diff
@@ -0,0 +1,106 @@
...
+        hamcrest = "2.2"
+        jackson = "2.10.2"
+        junit = "5.7.1"
+        logback = "1.2.3"
+        sl4j = "1.7.30"
+        springboot = "2.4.5"
+        springshell = "2.0.1.RELEASE"
+    }
+
+    //
```
wchamber: remove me?

stairctl/build.gradle
```diff
@@ -0,0 +1,106 @@
...
+
+    // Findbugs annotations, so we can selectively suppress findbugs findings
+    implementation group: 'com.google.code.findbugs', name: 'annotations', version: "${findbugsAnnotations}"
+
+    runtimeOnly group: "org.postgresql", name: "postgresql", version: "42.2.18"
+    runtimeOnly group: 'org.codehaus.groovy', name: 'groovy', version: '3.0.7'
+
+    testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api', version: "${junit}"
+    testImplementation group: 'org.hamcrest', name: 'hamcrest', version: "${hamcrest}"
+    //testImplementation group: 'org.springframework.boot', name: 'spring-boot-starter-test'
```
wchamber: remove me?

stairctl/build.gradle
```diff
@@ -0,0 +1,106 @@
...
+    }
+}
+
+plugins {
+    id 'java'
+    id 'application'
+    id 'com.github.spotbugs' version '4.0.8'
+    id "io.spring.dependency-management" version "1.0.10.RELEASE"
+    id "org.springframework.boot" version "2.4.0"
+    // Spotless doesn't work still
```
wchamber: optional  nit: consider filing a ticket for tracking, or just remove the commented out code.

stairctl/src/main/java/bio/terra/stairctl/StairwayService.java
```diff
@@ -0,0 +1,93 @@
...
+
+    try {
+      this.stairway =
+          Stairway.newBuilder()
+              .stairwayName("stairctl") // No queue, no flights, so name is irrelevant...
+              .stairwayClusterName("stairctlcluster") // ...as is clustername
+              .enableWorkQueue(false) // we will not be running any flights anywhere!
+              .build();
+
+      stairway.initialize(dataSource, false, false);
```
wchamber, line 47: might be nice to comment about how this is explicitly *not* trying to start the stairway instance and why.

stairctl/src/main/java/bio/terra/stairctl/commands/Output.java
```diff
@@ -0,0 +1,222 @@
...
+    List<ImmutablePair<String, String>> pairList = new ArrayList<>();
+    pairList.add(new ImmutablePair<>("dbname", connectParams.getDbname()));
+    pairList.add(new ImmutablePair<>("username", connectParams.getUsername()));
+    pairList.add(new ImmutablePair<>("host", connectParams.getHost()));
+    pairList.add(new ImmutablePair<>("port", connectParams.getPort()));
+    System.out.println("Stairway Connection:");
+    display("  ", pairList);
+  }
+
+  private static int display(String indent, List<ImmutablePair<String, String>> pairList) {
```
wchamber, line 211: nit: a quick comment about how this is trying to format things would be helpful

stairctl/src/main/java/bio/terra/stairctl/commands/Output.java
```diff
@@ -0,0 +1,222 @@
...
+    return maxLength;
+  }
+
+  public static void error(String userMessage, Exception ex) {
+    System.err.println(": " + ex.getMessage());
+    logger.error(userMessage + ": ", ex);
+  }
+
+  private static String shortenClassName(String in) {
+    if (in.length() <= 30) {
```
wchamber: I think the class name includes the package. Instead of truncating the end of the class name, what if we removed or abbreviated the package name first? Maybe overkill.

wchamber (approved):
Cool!
Consider writing unit tests for `Control`. If we're going to use this to manage our production services, it's important that we know that it's doing what we think it is.

STAIRCTL.md
ddietterich: Sounds good. I added those to the futures list.

library/src/main/java/bio/terra/stairway/Control.java
ddietterich: Good idea

library/src/main/java/bio/terra/stairway/Control.java
ddietterich: I like it. Thanks!

stairctl/build.gradle
ddietterich: PF-867

stairctl/src/main/java/bio/terra/stairctl/commands/Output.java
ddietterich: I found an Apache commons method to do the abbreviation, so it will be like what we see in the logs.

pshapiro4broad:
> The Spring Shell puts you at its own interactive prompt. It is like running psql. So writing JSON output wouldn't be easily consumable. If there is a need for that, it could be done. We would need to add some command line processing (like -c in psql) that would connect, run a command, and then exit. I have read stackoverflow conversations about how to do that - it is possible.

Interesting, I've never worked with spring shell before, I'll have to read up on it. My reason for bringing it up is because so often we work with programs that don't produce parseable output, and as soon as you want to wrap automation around them you've got to write some ugly `awk`/`sed`/`grep` code to munge the output. But that can wait until the need arises, if it ever does.

Also, after thinking about how to use this with TDR, what we'd want is a web endpoint for this API. That way we can use the existing authz to handle the db connection permission, and the results would be JSON. Although if this is done, it would be nice to do it in a way that the endpoint code and openapi yaml is shared across users of stairway. Not sure how that would be done though.

ddietterich:
I'm not understanding your suggest, but am very interested in ideas for making Stairway tooling as useful as possible. Maybe we can discuss live.

PR #82, opened 2021-07-01
[PF-883] Rename library to stairway; use convention plug-ins

ddietterich:
The Stairway project will be getting two new libraries (gcp and azure), so I wanted to rearrange the code. I took a cue from how @jmczerk did things in CRL to set up these convention plugins. The great news is that with the convention plugins, Spotless is set up identically in all projects and no longer complains.

I cannot test publishing without actually publishing, so I will debug that once this much has been reviewed. The **only** change here is in the build. Since I did a rename, there are (once again) a slew of renamed files that you can ignore. The reason for renaming is so I can use the common publish creation code (stolen shamelessly from John), to setup the publishing objects. Here are the files to look at:
```
build.gradle
buildSrc/build.gradle
buildSrc/src/main/groovy/stairway.java-conventions.gradle
buildSrc/src/main/groovy/stairway.library-conventions.gradle
settings.gradle
stairctl/build.gradle
stairctl/src/main/java/bio/terra/stairctl/StairwayService.java
stairctl/src/main/java/bio/terra/stairctl/commands/ConnectCommands.java
stairctl/src/main/java/bio/terra/stairctl/commands/Output.java
stairctl/src/main/java/bio/terra/stairctl/commands/StairwayCommands.java
stairway/build.gradle
```


wchamber (approved):
LGTM

jmczerk (approved):
Looks good, I like seeing the "library conventions" inheriting "java conventions" in action.

PR #85, opened 2021-07-26
TypeReference experiment.

jmczerk:
PR to experiment/discuss [this PR thread](https://github.com/DataBiosphere/stairway/pull/74#discussion_r675127129).

CC: @pshapiro4broad , @ddietterich

### DIFF
--- a/stairway/src/main/java/bio/terra/stairway/FlightMap.java
+++ b/stairway/src/main/java/bio/terra/stairway/FlightMap.java
@@ -7,6 +7,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
--- a/stairway/src/main/java/bio/terra/stairway/FlightMap.java
+++ b/stairway/src/main/java/bio/terra/stairway/FlightMap.java
@@ -7,7 +7,6 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import java.io.IOException;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -92,6 +91,26 @@ public class FlightMap {
   /** Check map for emptiness */
   boolean isEmpty() {
     return map.isEmpty();
+  }
+
+  @Nullable
+  public <T> T get(String key, TypeReference<T> typeReference) {
+    String value = map.get(key);
+
+    if (value == null) {
+      return null;
+    }
+
+    try {
+      return getObjectMapper().readValue(value, typeReference);
+    } catch (JsonProcessingException ex) {
+      throw new JsonConversionException(
+          "Failed to deserialize value '"
+              + value
+              + "' from JSON to type "
+              + typeReference.getType().getTypeName(),
+          ex);
+    }
   }
 
   /**
--- a/stairway/src/test/java/bio/terra/stairway/FlightMapTest.java
+++ b/stairway/src/test/java/bio/terra/stairway/FlightMapTest.java
@@ -2,9 +2,13 @@ package bio.terra.stairway;
 
 import bio.terra.stairway.exception.JsonConversionException;
 import bio.terra.stairway.fixtures.FlightsTestPojo;
+import com.fasterxml.jackson.core.type.TypeReference;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.UUID;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Tag;
@@ -53,16 +57,31 @@ public class FlightMapTest {
     FlightsTestPojo pojoOut = inMap.get(pojoKey, FlightsTestPojo.class);
     Assertions.assertEquals(pojoIn, pojoOut);
 
+    pojoOut = inMap.get(pojoKey, new TypeReference<>() {});
+    Assertions.assertEquals(pojoIn, pojoOut);
+
     Integer intOut = inMap.get(intKey, Integer.class);
+    Assertions.assertEquals(intIn, intOut);
+
+    intOut = inMap.get(intKey, new TypeReference<>() {});
     Assertions.assertEquals(intIn, intOut);
 
     String stringOut = inMap.get(stringKey, String.class);
     Assertions.assertEquals(stringIn, stringOut);
 
+    stringOut = inMap.get(stringKey, new TypeReference<>() {});
+    Assertions.assertEquals(stringIn, stringOut);
+
     UUID uuidOut = inMap.get(uuidKey, UUID.class);
     Assertions.assertEquals(uuidIn, uuidOut);
 
+    uuidOut = inMap.get(uuidKey, new TypeReference<>() {});
+    Assertions.assertEquals(uuidIn, uuidOut);
+
     MyEnum enumOut = inMap.get(enumKey, MyEnum.class);
+    Assertions.assertEquals(enumOut, enumIn);
+
+    enumOut = inMap.get(enumKey, new TypeReference<>() {});
     Assertions.assertEquals(enumOut, enumIn);
   }
 
@@ -152,5 +171,29 @@ public class FlightMapTest {
     // Deserializing map from bad JSON throws
     Assertions.assertThrows(
         JsonConversionException.class, () -> FlightMap.create(new ArrayList<>(), "garbage"));
+  }
+
+  @Test
+  public void typeReferences() {
+    FlightMap flightMap = new FlightMap();
+
+    Map<String, List<UUID>> inputMap = new HashMap<>();
+
+    inputMap.put(
+        "mapKey",
+        new ArrayList<>(
+            Arrays.asList(
+                UUID.randomUUID(), UUID.randomUUID(), UUID.randomUUID(), UUID.randomUUID())));
+
+    final String key = "myList";
+    flightMap.put(key, inputMap);
+    System.out.println(flightMap.getRaw(key));
+
+    Map<String, List<UUID>> outputMap = flightMap.get(key, Map.class);
+    Assertions.assertEquals(inputMap, outputMap);
+
+    HashMap<String, ArrayList<UUID>> explicitOutputMap =
+        flightMap.get(key, new TypeReference<>() {});
+    Assertions.assertEquals(inputMap, explicitOutputMap);
   }
 }


stairway/src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -94,6 +93,26 @@ boolean isEmpty() {
     return map.isEmpty();
   }
```
jmczerk, line 96: It would probably be preferable to write one of the `get()` overloads in terms of one another, but I did not do that for this POC (it's a little messy, would pursue if we decided to implement).

stairway/src/test/java/bio/terra/stairway/FlightMapTest.java
```diff
@@ -53,17 +57,32 @@ private void verifyMap(FlightMap inMap) {
     FlightsTestPojo pojoOut = inMap.get(pojoKey, FlightsTestPojo.class);
     Assertions.assertEquals(pojoIn, pojoOut);
```
jmczerk, line 59: All of the tests that call `verifyMap()` pass.

jmczerk:
I think we could return `Optional<T>` for the new overload if we wanted to.  This does bring up an orthogonal (but kind of related) thought...  

In PF-827 we realized that there are places where we store a NULL value under a key in `FlightMap`.  I think the implications are the same for both a NULL return and an empty `Optional<T>`, but perhaps we need to add a `containsKey()` method for whoever is placing those NULLs?

stairway/src/test/java/bio/terra/stairway/FlightMapTest.java
```diff
@@ -153,4 +172,28 @@ public void serDesExceptions() {
...
+    inputMap.put(
+        "mapKey",
+        new ArrayList<>(
+            Arrays.asList(
+                UUID.randomUUID(), UUID.randomUUID(), UUID.randomUUID(), UUID.randomUUID())));
+
+    final String key = "myList";
+    flightMap.put(key, inputMap);
+    System.out.println(flightMap.getRaw(key));
+
```
jmczerk, line 191: From my reading of `TypeReference`, I expected that the `Object` overload portion of the test to fail here, and the `TypeReference<>` caller to pass.  However, I got the opposite result:

```
Failed to deserialize value '["java.util.HashMap",{"mapKey":["java.util.ArrayList",[["java.util.UUID","0410a8d5-ae13-41d8-8c5c-2934c81cf1cd"],["java.util.UUID","b49103d0-129e-4c35-8af9-2e83e086f220"],["java.util.UUID","78bab861-d097-46a9-b6c4-47ea8e0842e7"],["java.util.UUID","de256be6-89c2-45c3-a62f-bc4f4579990f"]]]}]' from JSON to type java.util.HashMap<java.lang.String, java.util.ArrayList<java.util.UUID>>
bio.terra.stairway.exception.JsonConversionException: Failed to deserialize value '["java.util.HashMap",{"mapKey":["java.util.ArrayList",[["java.util.UUID","0410a8d5-ae13-41d8-8c5c-2934c81cf1cd"],["java.util.UUID","b49103d0-129e-4c35-8af9-2e83e086f220"],["java.util.UUID","78bab861-d097-46a9-b6c4-47ea8e0842e7"],["java.util.UUID","de256be6-89c2-45c3-a62f-bc4f4579990f"]]]}]' from JSON to type java.util.HashMap<java.lang.String, java.util.ArrayList<java.util.UUID>>
	at bio.terra.stairway.FlightMap.get(FlightMap.java:106)
	at bio.terra.stairway.FlightMapTest.typeReferences(FlightMapTest.java:189)
        ...
```
This was surprising to me, but I suspect it has something to do with the fact that we use [`DefaultTyping`](https://fasterxml.github.io/jackson-databind/javadoc/2.9/com/fasterxml/jackson/databind/ObjectMapper.DefaultTyping.html), which preserves the type info needed to do the `Object` deser. I don't understand why that would hurt the `TypeReference<>` though -- perhaps a Jackson bug?

stairway/src/test/java/bio/terra/stairway/FlightMapTest.java
```diff
@@ -153,4 +172,28 @@ public void serDesExceptions() {
...
+  public void typeReferences() {
+    FlightMap flightMap = new FlightMap();
+
+    Map<String, List<UUID>> inputMap = new HashMap<>();
+
+    inputMap.put(
+        "mapKey",
+        new ArrayList<>(
+            Arrays.asList(
+                UUID.randomUUID(), UUID.randomUUID(), UUID.randomUUID(), UUID.randomUUID())));
```
pshapiro4broad, line 186: FYI the new collection APIs can streamline code like this
```suggestion
    Map<String, List<UUID>> inputMap =
      Map.of(
        "mapKey",
        List.of(UUID.randomUUID(), UUID.randomUUID(), UUID.randomUUID(), UUID.randomUUID()));
```

stairway/src/test/java/bio/terra/stairway/FlightMapTest.java
pshapiro4broad: I wonder if these types deserialize into different `Map` / `List` subclasses than the ones expected. Did it work if you assign the result of the TypeReference `get()` to `Map<String, List<UUID>>`? I'll have to try this myself I think.

ddietterich: So you are saying that line 195 failed (the error says line 189 which is a `put`) but you expected line 192 to fail?

What I have seen in the past is you get a compilation warning about the typing in the line 192 case, but it works.
Are you certain that the TypeReference in line 195 is correctly getting the type?

jmczerk: > I wonder if these types deserialize into different `Map` / `List` subclasses than the ones expected. Did it work if you assign the result of the TypeReference `get()` to `Map<String, List<UUID>>`? I'll have to try this myself I think.

I did initially just assign it to `outputMap` with the same results; using the concrete types was just an attempt to finesse the typing to line up.

jmczerk: > So you are saying that line 195 failed (the error says line 189 which is a `put`) but you expected line 192 to fail?
> 
> What I have seen in the past is you get a compilation warning about the typing in the line 192 case, but it works.
> Are you certain that the TypeReference in line 195 is correctly getting the type?

Yes, sorry the output predated running `spotlessApply` so the line numbers shifted.  Here is output with current code:
```Failed to deserialize value '["java.util.HashMap",{"mapKey":["java.util.ArrayList",[["java.util.UUID","aeeb06f7-2aff-4b14-b8cd-71fa409a5a01"],["java.util.UUID","610096dc-4791-4e75-ae80-c8634b5ed451"],["java.util.UUID","bb4143c8-1226-4ee9-87f3-5f58022b1274"],["java.util.UUID","584722d6-94cd-4d3c-822b-957a528c34c5"]]]}]' from JSON to type java.util.HashMap<java.lang.String, java.util.ArrayList<java.util.UUID>>
bio.terra.stairway.exception.JsonConversionException: Failed to deserialize value '["java.util.HashMap",{"mapKey":["java.util.ArrayList",[["java.util.UUID","aeeb06f7-2aff-4b14-b8cd-71fa409a5a01"],["java.util.UUID","610096dc-4791-4e75-ae80-c8634b5ed451"],["java.util.UUID","bb4143c8-1226-4ee9-87f3-5f58022b1274"],["java.util.UUID","584722d6-94cd-4d3c-822b-957a528c34c5"]]]}]' from JSON to type java.util.HashMap<java.lang.String, java.util.ArrayList<java.util.UUID>>
	at bio.terra.stairway.FlightMap.get(FlightMap.java:112)
	at bio.terra.stairway.FlightMapTest.typeReferences(FlightMapTest.java:196)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	...
```

Also using explicit typing on `TypeReference<>` instead of type inference does not make a difference.

pshapiro4broad: FWIW when I create a test program using the default ObjectMapper:
- the result of `get(String, Class<T>)` fails the assertion test - because it doesn't correctly map `String` to `UUID`, the `List` ends up being `List<String>` instead of `List<UUID>`
- the TypeReference code doesn't throw and passes the assert

[Main.java.txt](https://github.com/DataBiosphere/stairway/files/6880479/Main.java.txt)

jmczerk: > * the result of `get(String, Class<T>)` fails the assertion test - because it doesn't correctly map `String` to `UUID`, the `List` ends up being `List<String>` instead of `List<UUID>`

That doesn't surprise me, given that the type info included by `DefaultTyping` will not exist with the default mapper.  

It does surprise me that the `TypeReference` code doesn't seem to properly handle the embedded type info when the mapper uses `DefaultTyping` (which is what the error message leads me to believe is happening).

pshapiro4broad: I'll look a bit more. If we can't figure it out, one way around it is to use two different object mapper configs.

pshapiro4broad: Found a closed issue that covers this, https://github.com/FasterXML/jackson-databind/issues/2095

TLDR: if you don't use a complete type when serializing, you can't use a complete type when deserializing. Either using `objectMapper.writerFor(typeReference).writeValueAsString(value);` or
```
        Map<String, List<?>> explicitOutputMap = flightMap.get(key, new TypeReference<>() {});
```
fixes this. But I don't like either solution much for this code. Is `DefaultTyping.NON_FINAL` needed for the case that uses `TypeReference`? I don't understand what its function is in that case.

ddietterich: Do we know in advance that it is a `TypeReference` case? I don't see how we would.

stairway/src/test/java/bio/terra/stairway/FlightMapTest.java
```diff
@@ -153,4 +172,28 @@ public void serDesExceptions() {
...
+    flightMap.put(key, inputMap);
+    System.out.println(flightMap.getRaw(key));
+
+    Map<String, List<UUID>> outputMap = flightMap.get(key, Map.class);
+    Assertions.assertEquals(inputMap, outputMap);
+
+    HashMap<String, ArrayList<UUID>> explicitOutputMap =
+        flightMap.get(key, new TypeReference<>() {});
+    Assertions.assertEquals(inputMap, explicitOutputMap);
+  }
```
ddietterich, line 198: On a tangential thread - here at the bottom - the use of `DefaultTyping.NON_FINAL` is deprecated and is viewed as a security risk. It has something to do with a JSON insertion attack that would cause Jackson to execute arbitrary code in a constructor. I don't think we are very vulnerable to that attack: it would require access to the Stairway database to change the JSON. We are not accepting such forms over the wire from random clients.

If you control all of the classes that are serialized, you can fix this with a feature that restricts what can be constructed by a set of classes. For example, if everything was either a basic Java object or derived from StairwayObject. That seemed impractical at the layer where Stairway lives.

stairway/src/test/java/bio/terra/stairway/FlightMapTest.java
pshapiro4broad: > Do we know in advance that it is a TypeReference case? I don't see how we would.

We don't, which is why I was suggesting using a different `ObjectMapper` for the `get(String,TypeReference)` case. I think that would let us add this new API without changing existing functionality.

stairway/src/test/java/bio/terra/stairway/FlightMapTest.java
pshapiro4broad: Is `DefaultTyping.NON_FINAL` deprecated? The API we use is, but an equivalent form is not:
```
    ObjectMapper objectMapper = new ObjectMapper();
    objectMapper.activateDefaultTyping(objectMapper.getPolymorphicTypeValidator(),
            ObjectMapper.DefaultTyping.NON_FINAL, JsonTypeInfo.As.WRAPPER_ARRAY);
```

stairway/src/test/java/bio/terra/stairway/FlightMapTest.java
jmczerk: Well, I believe that we'd also need an additional `put()` overload that uses the alternate `ObjectMapper`, as they would not write compatible formats.

I personally would like to see Stairway get out of the ser/des business entirely for these parameters and treat them as completely opaque.  The original PR's that this came from are steps in that direction, though we still have work to do to figure out what we would like the final end result to look like (and that has unfortunately fallen off my priority list ~to some degree~).

I had originally proposed API's that passed a Serialization/Deserialization interface to allow callers to use arbitrary object-to-String ser/des logic (including-but-not-limited-to their own external Jackson Object Mapper), but they might have been over-engineering a bit.  They could probably be boiled down to something like:
```
public <T> void put(String key, Function<T, String> serialize)
public <T> T get(String key, Function<String, T> deserialize)
```
Then again, one can accomplish the same with the newly added `raw` put/get methods (which I think was the key argument against the serdes interfaces proposal)...

I definitely think that this is a topic that we should revisit again when priorities allow, but for now I think we should close this PR, otherwise I'll need to park in favor of some other work.

ddietterich: I agree.

pshapiro4broad: > Well, I believe that we'd also need an additional put() overload that uses the alternate ObjectMapper, as they would not write compatible formats.

Based on my testing this wasn't the case. I used the existing `put()` and the new `get(String, TypeReference)` and it worked fine, at least for the cases I tried. Maybe there's a specific case you have in mind that wouldn't work? I'd be interested to try it out if so.

> I had originally proposed API's that passed a Serialization/Deserialization interface to allow callers to use arbitrary object-to-String ser/des logic (including-but-not-limited-to their own external Jackson Object Mapper), but they might have been over-engineering a bit.

I guess this wouldn't be too bad. There should be a way to make a `Function` subclass that uses type inference so it would end up matching the `TypeReference` form.

jmczerk: > Maybe there's a specific case you have in mind that wouldn't work? I'd be interested to try it out if so.

Having two differently configured Object Mappers does make me nervous on principle, but I don't have a concrete use case that breaks it off the top of my head.  It also adds more serialization logic and opaque Jackson behavior inside of Stairway, when I'd really rather see that logic moved out.  

Have we identified a case that adding the `TypeReference<>` overload would enable that we can't do with the existing Class method today?  If so it'd be good to capture via Jira ticket so we can prioritize making the change, with a link to this PR (which I intend to close later today).

pshapiro4broad: As I see it there are two benefits:
- users can retrieve generic typed data without requiring an unsafe type cast
- using an inferred type avoids code duplication

Given that it's not trivial to add, and the benefits aren't huge, it's probably not worth doing. Thanks for taking the time to look into it.

jmczerk: Thanks for your inputs too!  

FWIW, by removing any deserialization of JSON to `Object` type in [PF-703](https://github.com/DataBiosphere/stairway/pull/74) we aren't doing any more casts (at least not directly in our code), which resolves a bunch of issues that we are working around in several places related to such illegal casts; so we are in a better place than we were when we were before that PR in this area.

jmczerk:
Closing, can revisit in future.

PR #86, opened 2021-08-03
Add TypeReference<T> Overload and Register Guava Module with ObjectMapper

jmczerk:
So... turns out that the `TypeReference<T>` overload was needed after all, where Workspace Manager is storing `Map<enum, T>` in several places that break once we are no longer caching object references in the map...

Also adds the ability to ser/des Guava containers (also done in WS Manager, maybe only in tests...)

### DIFF
--- a/stairway/src/main/java/bio/terra/stairway/FlightMap.java
+++ b/stairway/src/main/java/bio/terra/stairway/FlightMap.java
@@ -95,13 +95,26 @@ public class FlightMap {
   }
 
   /**
-   * Return the object from the hash map deserialized to the right type. Throw an exception if the
-   * Object cannot be deserialized to that type.
+   * Returns true if this map contains a mapping of the passed key.
+   *
+   * @param key to look up in the map
+   * @return true if present
+   */
+  boolean containsKey(String key) {
+    return map.containsKey(key);
+  }
+
+  /**
+   * Return the object from the hash map deserialized to the right non-parameterized type. Throw an
+   * exception if the Object cannot be deserialized to that type. For parameterized types, the
+   * {@code TypeReference<T>} overload of this method offers more type safety and should be
+   * preferred.
    *
    * @param <T> - type of class to expect in the hash map
    * @param key - key to lookup in the hash map
    * @param type - class requested
-   * @return null if not found
+   * @return null if not found or if a null value is stored at that key (use method {@code
+   *     containsKey()} to differentiate)
    * @throws JsonConversionException if found, not deserializable to the requested type
    */
   @Nullable
@@ -117,6 +130,38 @@ public class FlightMap {
     } catch (JsonProcessingException ex) {
       throw new JsonConversionException(
           "Failed to deserialize value '" + value + "' from JSON to type " + type.getName(), ex);
+    }
+  }
+
+  /**
+   * Return the object from the hash map deserialized to the right type. Throw an exception if the
+   * Object cannot be deserialized to that type. This overload is preferred when deserializing
+   * parameterized types as it provides stronger type checking at deserialization time.
+   *
+   * @param <T> - type of class to expect in the hash map
+   * @param key - key to lookup in the hash map
+   * @param typeReference - class requested
+   * @return null if not found or if a null value is stored at that key (use method {@code
+   *     containsKey()} to differentiate)
+   * @throws JsonConversionException if found, not deserializable to the requested type
+   */
+  @Nullable
+  public <T> T get(String key, TypeReference<T> typeReference) {
+    String value = map.get(key);
+
+    if (value == null) {
+      return null;
+    }
+
+    try {
+      return getObjectMapper().readValue(value, typeReference);
+    } catch (JsonProcessingException ex) {
+      throw new JsonConversionException(
+          "Failed to deserialize value '"
+              + value
+              + "' from JSON to type "
+              + typeReference.getType().getTypeName(),
+          ex);
     }
   }
 
--- a/stairway/src/main/java/bio/terra/stairway/FlightMap.java
+++ b/stairway/src/main/java/bio/terra/stairway/FlightMap.java
@@ -90,7 +90,7 @@ public class FlightMap {
   }
 
   /** Check map for emptiness */
-  boolean isEmpty() {
+  public boolean isEmpty() {
     return map.isEmpty();
   }
 
@@ -100,22 +100,22 @@ public class FlightMap {
    * @param key to look up in the map
    * @return true if present
    */
-  boolean containsKey(String key) {
+  public boolean containsKey(String key) {
     return map.containsKey(key);
   }
 
   /**
-   * Return the object from the hash map deserialized to the right non-parameterized type. Throw an
-   * exception if the Object cannot be deserialized to that type. For parameterized types, the
+   * Return the object from the flight map deserialized to the right non-parameterized type. Throw
+   * an exception if the Object cannot be deserialized to that type. For parameterized types, the
    * {@code TypeReference<T>} overload of this method offers more type safety and should be
    * preferred.
    *
-   * @param <T> - type of class to expect in the hash map
-   * @param key - key to lookup in the hash map
+   * @param <T> - type of class to expect in the flight map
+   * @param key - key to lookup in the flight map
    * @param type - class requested
    * @return null if not found or if a null value is stored at that key (use method {@code
    *     containsKey()} to differentiate)
-   * @throws JsonConversionException if found, not deserializable to the requested type
+   * @throws JsonConversionException if not deserializable to the requested type
    */
   @Nullable
   public <T> T get(String key, Class<T> type) {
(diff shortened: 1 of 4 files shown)


stairway/src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -120,6 +133,38 @@ boolean isEmpty() {
...
   }
 
+  /**
+   * Return the object from the hash map deserialized to the right type. Throw an exception if the
+   * Object cannot be deserialized to that type. This overload is preferred when deserializing
+   * parameterized types as it provides stronger type checking at deserialization time.
+   *
+   * @param <T> - type of class to expect in the hash map
+   * @param key - key to lookup in the hash map
+   * @param typeReference - class requested
```
pshapiro4broad, line 143: It's a shame the api is fixed as this is great case for `Optional`

stairway/src/test/java/bio/terra/stairway/FlightMapTest.java
```diff
@@ -153,4 +188,48 @@ public void serDesExceptions() {
     Assertions.assertThrows(
         JsonConversionException.class, () -> FlightMap.create(new ArrayList<>(), "garbage"));
   }
+
+  @Test
+  public void guavaTypes() {
+    FlightMap flightMap = new FlightMap();
+
+    final String key = "immutableList";
+    ImmutableList<Integer> immutableListIn = ImmutableList.of(1, 2, 3, 4);
```
pshapiro4broad, line 197: I'm curious as to why you'd use a guava type for an immutable list rather than use a JDK type.

The guava classes were great when the JDK was lacking features, but now that the JDK has (somewhat) caught up I think it's better to avoid adding a dependency if possible.

stairway/src/test/java/bio/terra/stairway/FlightMapTest.java
```diff
@@ -153,4 +188,48 @@ public void serDesExceptions() {
...
+  @Test
+  public void typeRef() {
+    FlightMap flightMap = new FlightMap();
+    final String key = "key";
+
+    Map<MyEnum, FlightsTestPojo> testMapIn = new HashMap<>();
+    testMapIn.put(MyEnum.FOO, new FlightsTestPojo().anint(1).astring("first"));
+    testMapIn.put(MyEnum.BAR, new FlightsTestPojo().anint(2).astring("second"));
+    testMapIn.put(MyEnum.BAZ, new FlightsTestPojo().anint(3).astring("third"));
+    flightMap.put(key, testMapIn);
```
pshapiro4broad, line 213: Does `Map.of` not work here? That would be a slightly easier way to write this.

stairway/src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -120,6 +133,38 @@ boolean isEmpty() {
...
+   * @return null if not found or if a null value is stored at that key (use method {@code
+   *     containsKey()} to differentiate)
+   * @throws JsonConversionException if found, not deserializable to the requested type
```
pshapiro4broad: might be clearer to say
```suggestion
   * @throws JsonConversionException if found but not deserializable to the requested type
```
Or even
```suggestion
   * @throws JsonConversionException if not deserializable to the requested type
```
Since the missing case is already documented.

stairway/src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -95,13 +95,26 @@ boolean isEmpty() {
...
 
   /**
-   * Return the object from the hash map deserialized to the right type. Throw an exception if the
-   * Object cannot be deserialized to that type.
+   * Returns true if this map contains a mapping of the passed key.
+   *
+   * @param key to look up in the map
+   * @return true if present
+   */
+  boolean containsKey(String key) {
```
ddietterich: Should we make this public? I think we can't distinguish null value from not-found in the FlightMap right now.

stairway/src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -95,13 +95,26 @@ boolean isEmpty() {
...
+    return map.containsKey(key);
+  }
+
+  /**
+   * Return the object from the hash map deserialized to the right non-parameterized type. Throw an
```
pshapiro4broad: Personally I would avoid the term `hash map` here as it's an implementation detail that the caller doesn't need to know about. If we changed `FlightMap` to use a different underlying data structure it shouldn't affect the api behavior. I would say something like `Given a key, return the flight map object for that key.`

pshapiro4broad (approved):
Well, this was a nice turn of events! I was just looking at the TDR code's use of `FlightMap` and there are many cases where a generic type is stored and retrieved.

ddietterich (approved):
LGTM - one suggestion

stairway/src/main/java/bio/terra/stairway/FlightMap.java
pshapiro4broad: I was thinking that maybe a missing key should be an error instead of returning `null`. It seems reasonable to expect users to ensure that a key is always set before it's retrieved. If they want to handle the unset case, they can always catch the `KeyNotFound` exception.

ddietterich: My original aim was to keep the interface compatible with Map. We are a bit afield of that now, so we could consider this. It would not be upward compatible.

pshapiro4broad: You should run it by more people than just me, but I think it's better to throw than return `null`. The flight map is used to pass data between steps, and I think it should be an error for a step to depend on a key that isn't always present.

stairway/src/test/java/bio/terra/stairway/FlightMapTest.java
jmczerk: Yeah, Dan and I were on the fence about this.  There were a few places in WSM that were using `ImmutableList` in what appeared to be test code only, I leaned towards keeping it in as a hedge so as not to have to rev again if that turned out to be wrong (the test-only assumption).  I could be convinced to back it out though.

stairway/src/main/java/bio/terra/stairway/FlightMap.java
jmczerk: > Should we make this public? I think we can't distinguish null value from not-found in the FlightMap right now.

Good catch, this was meant to be public.

> You should run it by more people than just me, but I think it's better to throw than return null.

This sounds reasonable to me, but I'd rather not lump it into this effort, as it's already turned up a few more unexpected behaviors than we would have liked...

stairway/src/main/java/bio/terra/stairway/FlightMap.java
jmczerk: Given this is a new call we could make this `Optional` if we don't care about mixing return types on overloads.  @ddietterich what do you think?

pshapiro4broad: Does Java let you change the return type of an overloaded method? Even if it does, I think that would be unexpected for users if the api. Instead I think you'd have to deprecate `get` and add a new method.

stairway/src/test/java/bio/terra/stairway/FlightMapTest.java
pshapiro4broad: I don't think the TDR code will need this support, but I can understand stairway wanting to support other possible use cases. When custom serializer support is added this will no longer be your problem. 😀

ddietterich: I think the addition of guava is safe and defends against any existing guava use.

With the recent addition of `getRaw` and `putRaw` you can now serialize/deserialize to your heart's content without Stairway messing with your data.

stairway/src/test/java/bio/terra/stairway/FlightMapTest.java
jmczerk: Weird, when I replace the repeated push with `Map.of`:
```
    Map<MyEnum, FlightsTestPojo> testMapIn = Map.of(
      MyEnum.FOO, new FlightsTestPojo().anint(1).astring("first"),
      MyEnum.BAR, new FlightsTestPojo().anint(2).astring("second"),
      MyEnum.BAZ, new FlightsTestPojo().anint(3).astring("third")
    );

```
I get different JSON that doesn't play nice with either deserializer:
```
{"FOO":["bio.terra.stairway.fixtures.FlightsTestPojo",{"astring":"first","anint":1}],"BAZ":["bio.terra.stairway.fixtures.FlightsTestPojo",{"astring":"third","anint":3}],"BAR":["bio.terra.stairway.fixtures.FlightsTestPojo",{"astring":"second","anint":2}]}
```
With the code as-is I get:
```
["java.util.HashMap",{"BAR":["bio.terra.stairway.fixtures.FlightsTestPojo",{"astring":"second","anint":2}],"FOO":["bio.terra.stairway.fixtures.FlightsTestPojo",{"astring":"first","anint":1}],"BAZ":["bio.terra.stairway.fixtures.FlightsTestPojo",{"astring":"third","anint":3}]}]
```
Given the `TypeReference<T>` is working for the WSM case, I'm inclined not to dig too deep...

stairway/src/main/java/bio/terra/stairway/FlightMap.java
jmczerk: Fixed on both overloads.

stairway/src/main/java/bio/terra/stairway/FlightMap.java
jmczerk: Good suggestion, replaced throughout.

stairway/src/test/java/bio/terra/stairway/FlightMapTest.java
pshapiro4broad: That's unfortunate. I would much rather see support of JDK types like `Map.of()` than guava types in stairway, although I don't know what's needed to fix this. TDR may end up using `Map.of()` in the future so I guess we'll figure it out then.

PR #87, opened 2021-08-17
[PF-884] Stairway refactor for multi-cloud

ddietterich:
This PR implements a code restructuring in Stairway. The changes are motivated and described in [Stairway Code Restructuring](https://docs.google.com/document/d/1j6tf4YByB1zoDarVYL5Avd65JpFRMrkH7bdc4LvP1bY/edit#)

In addition, it changes the library structure, providing a base library with no queue implementation (stairway) and a GCP library with the existing PubSub queue implementation (stairway-gcp). The GCP library exposes the stairway library interface.

The next step after getting this merged and published will be to modify Terra Common Library to use the new method of constructing Stairway and depend on stairway-gcp. And then roll it out through the various components.


stairway/src/main/java/bio/terra/stairway/QueueInterface.java
```diff
@@ -0,0 +1,32 @@
+package bio.terra.stairway;
+
+public interface QueueInterface {
+
+  /**
+   * Read some messages from the queue and request that Stairway process each one.
+   *
+   * @param dispatchContext anonymous context passed to the processFunction
```
wchamber: optional: The dispatchContext passing around feels a little funny to me. I think it happens because the QueueMessage has it's own process message behavior as a part of the class, so it needs the context method to actually do the processing. But if we invert that and leave the message processing to outside of the message itself, I think the interfaces and classes get simpler. Having the message be a dumb data record is what I expected going in.
QueueProcessFunction can have a `boolean process(String message)` signature.
The WorkQueueListener can take the `String` and give it to QueueMessage to parse. Then it can look at a QueueMessage::Type enum, or a visitor pattern on QueueMessage types and call `stairwayImpl.resume`.

stairway/src/main/java/bio/terra/stairway/QueueInterface.java
```diff
@@ -0,0 +1,32 @@
...
+      Object dispatchContext, int maxMessages, QueueProcessFunction processFunction)
+      throws InterruptedException;
+
+  /**
+   * Put a message on the queue.
+   *
+   * @param message the message to enqueue
+   * @throws InterruptedException allows for queue waits to throw if the calling thread is shutdown
+   */
+  void enqueueMessage(String message) throws InterruptedException;
```
wchamber: optional nit: consider making the name of `dispatchMessages` and `enqueueMessages` parallel verbs, like `dequeue` and `enqueue` or `push` and `pop`/`pull`.

stairway/src/main/java/bio/terra/stairway/QueueInterface.java
```diff
@@ -0,0 +1,32 @@
...
+
+  /**
+   * Remove all messages from a queue. NOTE: This method is intended for controlled test
+   * environments where a cloud platform queue is reused. It should clear all messages from the
+   * queue, simply ignoring them, so that tests can start from an empty queue.
+   */
+  void purgeQueue();
```
wchamber: consider adding a scary word to the name to make it less likely that this will be used outside of testing. e.g. `unsafe`, `dangerous`, or `forTest`

stairway-gcp/src/test/java/bio/terra/stairway/gcp/PubSubQueueTest.java
```diff
@@ -0,0 +1,95 @@
...
+  @AfterEach
+  public void teardown() throws Exception {
+    workQueue.shutdown();
+    GcpQueueUtils.deleteQueue(projectId, topicId, subscriptionId);
+  }
+
+  private Map<String, Boolean> messages;
+
+  @Test
+  public void simpleTest() throws Exception {
```
wchamber: this looks a lot like what's in FileQueueTest. Consider making a test helper that takes a QueueInterface and runs these simple tests on its implementation. Then if we change something about how we expect our interface to change or add new tests, all of our implementation tests can be changed in a single place.

stairway-gcp/src/test/java/bio/terra/stairway/gcp/PubSubQueueTest.java
```diff
@@ -0,0 +1,95 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Test to make sure the GcpPubSubQueue is basically working */
+@Tag("connected")
+public class PubSubQueueTest {
```
wchamber: style nit: name the test class `GcpPubSubQueueTest` since it's the GcpPubSubQueue that's being tested. https://google.github.io/styleguide/javaguide.html#s5.2.2-class-names

stairway-gcp/src/test/java/bio/terra/stairway/gcp/QueueBuilderTest.java
```diff
@@ -0,0 +1,44 @@
...
+import bio.terra.stairway.gcp.GcpPubSubQueue.Builder;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+@Tag("unit")
+public class QueueBuilderTest {
+
+  @Test
+  public void missingArguments() {
```
wchamber, line 12: style nit: include this in the same file as the rest of the test for GcpPubSubQueue so that the tests directly on that class are all in the same place.

stairway/src/main/java/bio/terra/stairway/impl/ControlImpl.java
```diff
@@ -0,0 +1,290 @@
...
+import java.util.List;
+import java.util.UUID;
+import javax.sql.DataSource;
+
+/**
+ * This class provides an API for out-of-band debugging and recovering Stairway flights. It caters
+ * to the use cases of the stairctl tool. The methods are constrained to only access the database,
+ * and do not rely on any application-specific state. We current do not use the FlightDao for the
+ * database accesses. We may be able to do so once FlightDao stops doing deserialization, so we take
+ * a reference to the flightDao even though it is unused.
```
wchamber: nit: this seems to duplicate the comment on `Control`. Let's have a single source of correctness for this comment. We could say `see {@link Control}` here.

stairway/src/main/java/bio/terra/stairway/Control.java
```diff
@@ -1,307 +1,48 @@
...
-
-    try (var connection = dataSource.getConnection();
-        var statement = new NamedParameterPreparedStatement(connection, sb.toString())) {
-      DbUtils.startReadOnlyTransaction(connection);
-      List<Control.Flight> flightList = flightQuery(statement, offset, limit);
-      DbUtils.commitTransaction(connection);
-      return flightList;
-    }
-  }
+public interface Control {
```
wchamber, line 19: optional: I'm not sure I understand why we're making this an interface, but I'm also not sure it matters very much. Is this only to "hide" the implementation from clients more, or do we expect multiple implementations? Naively I would think every public method on the Control impl is public for clients.

stairway/src/main/java/bio/terra/stairway/Flight.java
```diff
@@ -4,21 +4,21 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * Manage the atomic execution of a series of Steps This base class has the mechanisms for executing
- * the series of steps.
+ * Manage the atomic execution of a series of Steps
+ *
+ * <p>This base class has the mechanisms for executing the series of steps, retrying, maintaining
+ * state in a database,
```
wchamber: this says `base class` but is Flight ever extended?

stairway/src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -22,74 +19,41 @@
...
 
+  /** Construct an empty flight map */
   public FlightMap() {
     map = new HashMap<>();
   }
 
   /**
-   * Alternate constructor, used by the DAO to re-create FlightMap from its serialized form.
+   * This constructor is used when Stairway re-creates the flight map from the database. It is not
+   * intended for client use.
```
wchamber: public constructors that shouldn't be public make me a little nervous. Is there a reason clients should not create input FlightMaps with a Map<String, String> instead of `putRaw` repeatedly?
We could add a public putAllRaw(Map<String, String> values) method and use that in stairwayimpl as another way to keep to only having public methods we actually are ok with being public.

stairway/src/main/java/bio/terra/stairway/FlightSupport.java
```diff
@@ -0,0 +1,36 @@
+package bio.terra.stairway;
+
+import bio.terra.stairway.exception.StairwayException;
+
+/**
+ * Flight support interface
+ *
+ * <p>This interface provides additional methods for the {@link Flight} implementation. It lets us
+ * better separate the internal implementation of flights and the Stairway object rather than have
+ * them as public members referenced directly from the Flight object.
```
wchamber: I'm having a hard time time figuring out what I would or wouldn't add to this interface. Is there a better name for its purpose than `Support` or is it actually doing several different things that might be better as separate interfaces?
If it's useful to bundle these interfaces together, we could still have a concrete `Support` class which then has a single purpose of collecting what's needed by the Flight.
or per other comment, a `FlightImpl` class might make this class unnecessary.

stairway/src/main/java/bio/terra/stairway/QueueInterface.java
```diff
@@ -0,0 +1,32 @@
+package bio.terra.stairway;
+
+public interface QueueInterface {
```
wchamber, line 16: nit: add a class comment on what this is used for and why we might have different impls.

stairway/src/main/java/bio/terra/stairway/QueueProcessFunction.java
```diff
@@ -1,6 +1,13 @@
...
+ * When Stairway calls the queue implementation to dispatch messages, it provides a function with
+ * this signature. This function will attempt to activate the flight mentioned in the message. If
+ * that succeeds, this function returns true and the message can be consumed. If that fails, this
+ * function returns false and the message should remain on the queue. The typical reason for failure
+ * is that the instance of the service is shutting down and can no longer run flights.
+ */
 @FunctionalInterface
-public interface QueueProcessFunction<T1, T2, R> {
-  R apply(T1 t1, T2 t2) throws InterruptedException;
+public interface QueueProcessFunction {
```
wchamber, line 11: consider making this a nested interface of QueueInterface since it is really a part of the QueueInterface interface. i.e. `interface QueueInterface { .... interface ProcessFunction { ...} ...}`

stairway/src/main/java/bio/terra/stairway/exception/FlightControlException.java
```diff
@@ -0,0 +1,15 @@
+package bio.terra.stairway.exception;
+
+public abstract class FlightControlException extends StairwayException {
```
wchamber, line 7: add javadoc class comments for this and other exception classes

stairway/src/main/java/bio/terra/stairway/impl/FlightFilterAccess.java
```diff
@@ -0,0 +1,233 @@
...
+import com.fasterxml.jackson.core.JsonProcessingException;
+import java.sql.SQLException;
+import java.time.Instant;
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ * A FlightFilterAccess is used to access FlightFilter members for generating the SQL queries
+ * applying the predicates.
+ */
+public class FlightFilterAccess {
```
wchamber: if all the methods are package private, should the class still be public?

stairway/src/main/java/bio/terra/stairway/impl/FlightFilterAccess.java
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  // -- predicate methods --
+
+  /**
+   * Make the one SQL predicate for the flight table of this predicate.
+   *
+   * @return the SQL predicate
+   */
+  String makeFlightPredicateSql(FlightFilterPredicate predicate) {
```
wchamber, line 186: AFAICT, only `makeSql` and `storePredicateValues` should be non private. Or am I missing something about these other methods? If it's for testing, consider using `@VisibleForTesting` annotation.

stairway/src/main/java/bio/terra/stairway/impl/FlightFilterAccess.java
```diff
@@ -0,0 +1,233 @@
...
+import bio.terra.stairway.FlightFilter.FlightFilterPredicate;
+import bio.terra.stairway.StairwayMapper;
+import bio.terra.stairway.exception.FlightFilterException;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import java.sql.SQLException;
+import java.time.Instant;
+import org.apache.commons.lang3.StringUtils;
+
+/**
+ * A FlightFilterAccess is used to access FlightFilter members for generating the SQL queries
```
wchamber, line 14: nit: `Access` doesn't seem quite right to me. Maybe `QueryGenerator` or `SqlEngine`?

stairway/src/main/java/bio/terra/stairway/impl/StairwayImpl.java
```diff
@@ -0,0 +1,652 @@
...
+      throw new StairwayShutdownException("Stairway is shut down and cannot be initialized");
+    }
+
+    stairwayInstanceDao = new StairwayInstanceDao(dataSource);
+    flightDao =
+        new FlightDao(
+            dataSource, stairwayInstanceDao, exceptionSerializer, hookWrapper, stairwayName);
+    control = new ControlImpl(dataSource, flightDao, stairwayInstanceDao);
+    flightSupport = new FlightSupportImpl(this);
+    flightFactory = new FlightFactory();
```
wchamber: is there a reason these are instantiated here instead of the constructor? being here implies to me they need the dataSource somehow.

stairway/src/main/java/bio/terra/stairway/impl/StairwayImpl.java
```diff
@@ -0,0 +1,652 @@
...
+    if (retentionCheckInterval != null && completedFlightRetention != null) {
+      scheduledPool.scheduleWithFixedDelay(
+          new CompletedFlightCleaner(completedFlightRetention, flightDao),
+          retentionCheckInterval.toSeconds(),
+          retentionCheckInterval.toSeconds(),
+          TimeUnit.SECONDS);
+    }
+  }
+
+  private void startWorkQueueListener() {}
```
wchamber: remove me?

stairway/src/main/java/bio/terra/stairway/Flight.java
```diff
@@ -97,13 +87,9 @@ protected void addStep(Step step, RetryRule retryRule) {
    */
   public void run() {
     try {
-      hookWrapper().startFlight(flightContext);
+      flightSupport.startFlightHook(flightContext);
```
wchamber: What if we made a client facing Flight class/interface and ported all of the code originally in this file to a `stairwayimpl/FlightImpl` class? I think that class could refer to `StairwayImpl` directly and we could get rid of `FlightSupport` entirely. As-is, FlightSupport seems like a strange bit of indirection where Flight still knows a lot about bits of the internal implementation, but doesn't get access to the impl directly.

stairway/src/test/java/bio/terra/stairway/ScenarioTest.java
```diff
@@ -242,9 +242,15 @@ public void testQuietDown() throws Exception {
     String stairwayName = stairway.getStairwayName();
 
     stairway.terminate(5, TimeUnit.SECONDS);
-    stairway = null;
-
-    stairway = TestUtil.makeStairwayValidateRecovery(stairwayName, flightId, 1);
+    // out with the old; in with the new
```
wchamber: nit: this comment isn't very clear to me. ditto it's copy

stairway/src/test/java/bio/terra/stairway/fixtures/FileQueue.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.concurrent.TimeUnit;
+import org.apache.commons.io.FileUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class implements the QueueInterface as a temp directory on the local file system. It is only
+ * intended for testing.
```
wchamber, line 21: nit: probably doesn't matter now, but consider implementing the test queue in memory. Might be simpler with `java.util.Queue`. A single `java.util.Queue`  instance could be referenced by QueueInterface instances, like the shared `File queueDir`.

stairway/src/main/java/bio/terra/stairway/queue/QueueMessageReady.java
```diff
@@ -19,12 +20,13 @@ public QueueMessageReady(String flightId) {
...
     try {
-      boolean resumed = stairway.resume(flightId);
+      boolean resumed = stairwayImpl.resume(flightId);
       logger.info(
           "Stairway "
-              + stairway.getStairwayName()
+              + stairwayImpl.getStairwayName()
               + (resumed ? " resumed flight: " : " did not resume flight: ")
               + flightId);
     } catch (DatabaseOperationException ex) {
```
wchamber, line 41: should this propagate false back to the QueueProccessFunction if we get an exception? Do we know if the flightId has been resumed or not?

wchamber (commented):
I tried to restrict my review to new code, but I may have missed stuff.

settings.gradle
```diff
@@ -1,5 +1,6 @@
 rootProject.name = 'stairwaySaga'
 include 'stairway'
+include 'stairway-gcp'
 include 'stairctl'
 
-gradle.ext.version = "0.0.62-SNAPSHOT"
```
jmczerk, line 5: Is this a local test artifact, it seems like it would move the version backward?

stairway/src/main/java/bio/terra/stairway/impl/FlightMapUtils.java
```diff
@@ -0,0 +1,92 @@
...
+ * FlightMapUtils provides methods to create a flight map from data pulled from the database and
+ * format data from flight map to be stored in the database
+ */
+public class FlightMapUtils {
+  /**
+   * Create a flight map from an input list
+   *
+   * @param inputList input list form of the input parameters
+   */
+  static FlightMap makeFlightMap(List<FlightInput> inputList) {
```
jmczerk, line 25: I really like that you've moved the DAO/JSON handling stuff out of `FlightMap` proper.

Per Will's suggestion, you might consider having this just construct an empty `FlightMap` and `putRaw` into it during iteration, to avoid the new public ctor.

stairway/src/main/java/bio/terra/stairway/FlightMap.java
```diff
@@ -22,74 +19,41 @@
...
-    // If we have an empty list AND there is valid JSON, this predates the flightworking table...
-    // use the json in this case.
-    if (inputList.isEmpty() && json != null) {
-      FlightMap map = new FlightMap();
-      map.fromJson(json);
-      return map;
-    }
-
-    // Otherwise just use the input list (even if it's empty).
-    return new FlightMap(inputList);
```
jmczerk, line 84: Nit: if this is going to be public I'm a little nervous about returning a mutable reference to the internal map here.  Maybe consider calling it `asMap()` and have it return a copy (I assume we're doing DB ops when this is being used, so copy overhead would be negligible).

stairway/src/main/java/bio/terra/stairway/QueueInterface.java
ddietterich: Very helpful observation. I think I can simplify the interface as you suggest.

stairway-gcp/src/test/java/bio/terra/stairway/gcp/PubSubQueueTest.java
ddietterich: I am not crazy about the cut'n'paste. IIUC, I would have to generate another library just to hold the test helper and then use that in the testing of the other libraries. Seems like a lot of overhead. Is there something clever here that I am missing?

stairway/src/main/java/bio/terra/stairway/Flight.java
ddietterich: Another good idea. Yes, this is totally possible and a lot cleaner.

stairway/src/test/java/bio/terra/stairway/fixtures/FileQueue.java
ddietterich: Could have done. The advantage of files is it more closely models the semantics of PubSub: the messages survive the stop/start of Stairway instances.

stairway-gcp/src/test/java/bio/terra/stairway/gcp/PubSubQueueTest.java
wchamber: Ergh, I forgot that these were across different gradle projects. We would need to make a test library. The java-test-fixtures gradle plugin should make it possible to write shared test-only library code under `stairway/src/testFixtures/java....` and use it in `stairway/src/test` and `stairway-gcp/src/test`. CRL does this quite a bit e.g. [test lib code](https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/master/google-api-services-common/src/testFixtures/java/bio/terra/cloudres/google/api/services/common/testing/OperationTestUtils.java) and [dependency for test in other project](https://github.com/DataBiosphere/terra-cloud-resource-lib/blob/5cb870da728402388c1b86eae66ad506f9e2e8ef/google-compute/build.gradle#L14).

For reasons we didn't dig into, using that plugin did cause some spring boot issues last time we included it Stairway, but it's being used in CRL so there should be a non-harmful pattern.
Up to you on if the overhead is worth it. Maybe depends how many cloud specific interfaces we expect to have.

stairway/src/main/java/bio/terra/stairway/QueueInterface.java
ddietterich: I removed the `dispatchContext` and separated the message processor from the messages themselves. The `WorkQueueProcessor` implements the QueueProcessFunction. That function dispatches to the QueueMessage object.

stairway/src/main/java/bio/terra/stairway/QueueInterface.java
ddietterich: The reason I cannot do a simple `dequeue` interface is the requirement to ack or nack the message based on successful processing. That is the method PubSub has so that we avoid dropping messages on the floor in event of failure. They are not removed from the queue until they are ack'd. In this specific case, processing results in a successful database transaction. Stairway is set up to tolerate receiving the same message twice, but it will not tolerate losing a message in the queue.

stairway-gcp/src/test/java/bio/terra/stairway/gcp/PubSubQueueTest.java
ddietterich: I made [PF-958](https://broadworkbench.atlassian.net/browse/PF-958) to explore making test library code. I expect we will end up with two more queue libraries: Azure and AWS.

stairway/src/main/java/bio/terra/stairway/Control.java
ddietterich: The contract I'm trying to make is that what is public in the `stairway` package is the supported interface. What is in `impl` can change. I guess I need to be more explicit about that.

This separation exposes the Control interface as supported and allows the ControlImpl to use package-scoped methods in the `impl` package.

stairway/src/main/java/bio/terra/stairway/FlightSupport.java
ddietterich: You inspired a complete refactor of this code.

stairway/src/main/java/bio/terra/stairway/impl/StairwayImpl.java
ddietterich: FlightSupport is gone. I made FlightFactory static.

stairway/src/main/java/bio/terra/stairway/queue/QueueMessageReady.java
ddietterich: I added some documentation and made it return true. It is the not found case.

settings.gradle
ddietterich: Good catch. Merge mistake

stairway/src/main/java/bio/terra/stairway/impl/FlightMapUtils.java
ddietterich: Good suggestion. I have removed the constructor and switched to using putRaw.

stairway/src/main/java/bio/terra/stairway/FlightMap.java
ddietterich: Removed the extra constructor.

ddietterich: I still need `getMap` that exposes the map inside of the FlightMap. I don't see a good way around that.
I tried making a subclass that could then expose the map, but I can't cast to it from the FlightMap.
Any thoughts on how to get rid of it?

jmczerk: I was just thinking create an empty `Map<String, String>`, copy `map` contents into it, and return it, to avoid letting a caller accidentally mutate the `FlightMap` internal state (unless that is actually the intent?).

ddietterich: Even easier: I return `Collections.unmodifiableMap(map)` and don't have to copy the data.

ddietterich: Thanks for getting me on the right track!

jmczerk (approved):
Nicely done, and thanks for taking the feedback onboard!

stairway/src/main/java/bio/terra/stairway/Flight.java
```diff
@@ -1,426 +1,65 @@
...
-import org.apache.commons.lang3.builder.ToStringBuilder;
-import org.apache.commons.lang3.builder.ToStringStyle;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
- * Manage the atomic execution of a series of Steps This base class has the mechanisms for executing
- * the series of steps.
+ * The Flight object is the vehicle Stairway uses for (re)building the steps of a flight in
+ * preparation for running the flight. The object is not used in the actual execution of the flight.
```
wchamber: `The object is not used in the actual execution of the flight.` <- It's not clear to me what that means. Mind adding a little more?

stairway/src/main/java/bio/terra/stairway/Flight.java
```diff
@@ -1,426 +1,65 @@
...
-    }
-  }
-
-  private static final Logger logger = LoggerFactory.getLogger(Flight.class);
-
-  private final List<StepRetry> steps;
-  private final List<String> stepClassNames;
-  private FlightDao flightDao;
-  private FlightContext flightContext;
+public class Flight {
```
wchamber, line 15: no action: I like removing the execution behavior from this class a lot.

stairway/src/main/java/bio/terra/stairway/QueueInterface.java
```diff
@@ -0,0 +1,62 @@
...
+   * instances, cooperate the provide the service. In that configuration, Stairway uses a queue to
+   * share work across instances and to move work during shutdown. Stairway requires a transactional
+   * queue with at-least-once delivery semantics.
+   *
+   * <p>The queue interface is the abstraction Stairway uses to put messages onto the queue and read
+   * them from the queue for processing. At the time of this writing, the only message passed is the
+   * request for another Stairway instance to run a flight.
+   *
+   * <p>We need the queue abstraction to support multiple cloud platforms. On GCP, we use PubSub
+   * queues. On Azure, there are several messaging options.
```
wchamber: Thanks, these are great comments.
nit: move this part up as the class comment and leave the rest of this as the comment on `dispatchMessages`?

stairway/src/main/java/bio/terra/stairway/exception/FlightControlException.java
wchamber: gentle ping on comments for exception classes to make it clear when they should/should not be used :)

stairway/src/main/java/bio/terra/stairway/impl/FlightFactory.java
```diff
@@ -1,27 +1,23 @@
...
 import bio.terra.stairway.exception.MakeFlightException;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-// Default flight factory for Stairway
-public class StairwayFlightFactory implements FlightFactory {
-  Logger logger = LoggerFactory.getLogger(StairwayFlightFactory.class);
+class FlightFactory {
```
wchamber, line 12: nit: add a class comment.

stairway/src/main/java/bio/terra/stairway/queue/QueueMessage.java
```diff
@@ -0,0 +1,9 @@
+package bio.terra.stairway.queue;
+
+import bio.terra.stairway.impl.StairwayImpl;
+
+abstract class QueueMessage {
+  static final String FORMAT_VERSION = "1";
+
+  abstract boolean process(StairwayImpl stairway) throws InterruptedException;
```
wchamber, line 13: add comments for the class (and when it should be subclassed) as well as for `process`

stairway/src/main/java/bio/terra/stairway/queue/QueueMessageNoFields.java
```diff
@@ -16,5 +17,7 @@ public void setType(QueueMessageType type) {
   }
 
   @Override
-  public void process(Stairway stairway) {}
+  public boolean process(StairwayImpl stairwayImpl) {
+    return true;
```
wchamber: is this actually a trivial message that should return true, or should this message never actually be used to process? What does it mean for us to have a `QueueMessageNoFields` in the queue?
This does raise the question for me of if `QueueMessage` should even have behavior, or it's subclasses should just be dumb data holders. It seems reasonable for `WorkQueueProcessor` to do the actual `process` logic.

stairway/src/main/java/bio/terra/stairway/queue/QueueMessageReady.java
wchamber: thanks, the `return false` on catch was what I was asking about.

stairway/src/main/java/bio/terra/stairway/queue/QueueMessageReady.java
```diff
@@ -19,16 +20,20 @@ public QueueMessageReady(String flightId) {
...
     try {
-      boolean resumed = stairway.resume(flightId);
+      // Resumed is false if the flight is not found. We still call that a complete processing
+      // of the message and return true.
+      boolean resumed = stairwayImpl.resume(flightId);
       logger.info(
           "Stairway "
-              + stairway.getStairwayName()
+              + stairwayImpl.getStairwayName()
               + (resumed ? " resumed flight: " : " did not resume flight: ")
```
wchamber: I wouldn't have guessed from the log that "not resumed" means we didn't find the flight anymore and have given up on resuming it permanently. Maybe add some more of that to the log message?

wchamber (approved):
LGTM thanks!
Mostly just asking for comment tweaks and one last poke at QueueMessage.
(I'm assuming you didn't sneakily add large changes to the code that got pulled into new classes, otherwise let me know and I'll take another look)

stairway/src/main/java/bio/terra/stairway/queue/QueueMessageNoFields.java
ddietterich: The subclasses could be dumb data holders, and the `WorkQueueProcessor` could have a `switch` statement to direct processing. It is over-engineered and I would do it differently if I did it again. But my bias is to leave it be.

stairway/src/main/java/bio/terra/stairway/queue/QueueMessageReady.java
ddietterich: I re-worded the log message.

PR #89, opened 2021-09-08
[PF-988] Make GcpPubSubQueue public

ddietterich:
This class needs to be public so it is visible outside of stairway-gcp library.
Trivial fix.


wchamber (approved):

pshapiro4broad (approved):
LGTM. It would be nice to have javadoc for all public classes, though.

PR #91, opened 2022-01-05
PF-1272 Enhanced flight enumeration:

ddietterich:
This PR adds an enhanced flight enumeration endpoint and some bug fixes:
- validate filter inputs to avoid NPE. Previously, passing nulls for filter values would generate NPE.
- fix it so passing null to the complete_time properly filters for incomplete flights
- new getFlights endpoint that uses a nextPageToken in place of offset and returns total flights.
- additional tests to support the new filtering and endpoints

### DIFF
--- a/stairway/src/main/java/bio/terra/stairway/FlightEnumeration.java
+++ b/stairway/src/main/java/bio/terra/stairway/FlightEnumeration.java
@@ -1,0 +1,32 @@
+package bio.terra.stairway;
+
+import java.util.List;
+
+/** Class that holds the return value of the enhanced getFlights Stairway entrypoint */
+public class FlightEnumeration {
+  private final int totalFlights;
+  private final String nextPageToken;
+  private final List<FlightState> flightStateList;
+
+  public FlightEnumeration(
+      int totalFlights, String nextPageToken, List<FlightState> flightStateList) {
+    this.totalFlights = totalFlights;
+    this.nextPageToken = nextPageToken;
+    this.flightStateList = flightStateList;
+  }
+
+  /** @return total number of flights in the filtered set */
+  public int getTotalFlights() {
+    return totalFlights;
+  }
+
+  /** @return encoded string describing the start of the next page */
+  public String getNextPageToken() {
+    return nextPageToken;
+  }
+
+  /** @return list of FlightState in this page */
+  public List<FlightState> getFlightStateList() {
+    return flightStateList;
+  }
+}
--- a/stairway/src/main/java/bio/terra/stairway/FlightFilter.java
+++ b/stairway/src/main/java/bio/terra/stairway/FlightFilter.java
@@ -5,6 +5,7 @@ import bio.terra.stairway.exception.FlightFilterException;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
+import javax.annotation.Nullable;
 
 /**
  * A FlightFilter is used to filter the flights on a flight enumeration. You can build a filter with
@@ -14,7 +15,7 @@ import java.util.List;
  * FlightFilter filter = new FlightFilter()
  *     .addFilterSubmitTime(GREATER_THAN, yesterday)
  *     .addFilterFlightClass(EQUAL, IngestFlight.class)
- *     .addFilterFlightStatue(EQUAL, FlightStatus.COMPLETED)
+ *     .addFilterFlightStatus(EQUAL, FlightStatus.COMPLETED)
  *     .addFilterInputParameter("email", EQUAL, "ddtest@gmail.com");
  * }</pre>
  *
@@ -49,6 +50,9 @@ public class FlightFilter {
    * @return {@code this}, for fluent style
    */
   public FlightFilter addFilterSubmitTime(FlightFilterOp op, Instant timestamp) {
+    if (timestamp == null) {
+      throw new FlightFilterException("Submit filter timestamp cannot be null");
+    }
     FlightFilterPredicate predicate =
         new FlightFilterPredicate(
             "submit_time", op, timestamp, Datatype.TIMESTAMP, makeParameterName());
@@ -57,16 +61,21 @@ public class FlightFilter {
   }
 
   /**
-   * Filter by completed time. Note that completed time can be NULL.
+   * Filter by completed time. Passing null for the timestamp filters for incomplete flights
+   * regardless of the filter operand.
    *
    * @param op a {@link FlightFilterOp}
    * @param timestamp an Instant
    * @return {@code this}, for fluent style
    */
-  public FlightFilter addFilterCompletedTime(FlightFilterOp op, Instant timestamp) {
+  public FlightFilter addFilterCompletedTime(FlightFilterOp op, @Nullable Instant timestamp) {
     FlightFilterPredicate predicate =
         new FlightFilterPredicate(
-            "completed_time", op, timestamp, Datatype.TIMESTAMP, makeParameterName());
+            "completed_time",
+            op,
+            timestamp,
+            (timestamp == null ? Datatype.NULL : Datatype.TIMESTAMP),
+            makeParameterName());
     flightPredicates.add(predicate);
     return this;
   }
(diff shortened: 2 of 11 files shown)


stairway/src/main/java/bio/terra/stairway/FlightFilter.java
```diff
@@ -94,6 +118,9 @@ public FlightFilter addFilterFlightClass(FlightFilterOp op, Class<? extends Flig
    * @return {@code this}, for fluent style
    */
   public FlightFilter addFilterFlightStatus(FlightFilterOp op, FlightStatus status) {
+    if (status == null) {
+      throw new FlightFilterException("Status cannot be nul");
```
jmczerk, line 122: Nittiest of nits: `nul` -> `null`

jmczerk (approved):
Thanks for the detailed documentation!

stairway/src/test/java/bio/terra/stairway/impl/EnumerateFlightsTest.java
```diff
@@ -79,10 +81,15 @@ public void enumTest() throws Exception {
     flightList = flightDao.getFlights(0, 100, filter);
     checkResults("case 2", flightList, Arrays.asList("0", "1", "2", "3", "4", "5"));
 
-    // Case 3: date with null values - form1
+    // Case 3.1: date with values - form1
     filter = new FlightFilter().addFilterCompletedTime(FlightFilterOp.GREATER_THAN, minSubmit);
     flightList = flightDao.getFlights(0, 100, filter);
-    checkResults("case 3", flightList, Arrays.asList("0", "1", "2"));
+    checkResults("case 3.1", flightList, Arrays.asList("0", "1", "2"));
```
pshapiro4broad, line 87: FWIW we've been migrating our code from `Arrays.asList()` to `List.of()`, which is the "new" way to create lists.
```suggestion
    checkResults("case 3.1", flightList, List.of("0", "1", "2"));
```

stairway/src/main/java/bio/terra/stairway/impl/StairwayImpl.java
```diff
@@ -530,28 +531,24 @@ public FlightState getFlightState(String flightId)
...
+   * Implementation of getFlights - not for direct use. See {@link Stairway#getFlights(int, int,
+   * FlightFilter)}
    */
   public List<FlightState> getFlights(int offset, int limit, FlightFilter filter)
       throws StairwayException, DatabaseOperationException, InterruptedException {
     return flightDao.getFlights(offset, limit, filter);
   }
 
+  /**
+   * Implementation of getFlights - not for direct use. See {@link Stairway#getFlights(String,
```
pshapiro4broad, line 543: Not sure I understand the javadoc here. Isn't one of these internal and one external? Why do both javadocs `@link` back to themselves?

stairway/src/main/java/bio/terra/stairway/impl/StairwayImpl.java
```diff
@@ -530,28 +531,24 @@ public FlightState getFlightState(String flightId)
...
+   * Integer, FlightFilter)}
+   */
+  public FlightEnumeration getFlights(
```
pshapiro4broad, line 546: It's not required but I think it's helpful to use `@Override` on interface implementations like this.

stairway/src/main/java/bio/terra/stairway/FlightEnumeration.java
```diff
@@ -0,0 +1,32 @@
+package bio.terra.stairway;
+
+import java.util.List;
+
+/** Class that holds the return value of the enhanced getFlights Stairway entrypoint */
+public class FlightEnumeration {
+  private final int totalFlights;
+  private final String nextPageToken;
+  private final List<FlightState> flightStateList;
```
pshapiro4broad, line 9: It's a matter of taste but I find it redundant to include an object's type in its name. Plural names are more readable IMO, and don't need changing if the underling type changes.
```suggestion
  private final List<FlightState> flightStates;
```

stairway/src/main/java/bio/terra/stairway/FlightEnumeration.java
```diff
@@ -0,0 +1,32 @@
+package bio.terra.stairway;
+
+import java.util.List;
+
+/** Class that holds the return value of the enhanced getFlights Stairway entrypoint */
+public class FlightEnumeration {
```
pshapiro4broad, line 6: This doesn't look like an enumeration to me. It looks like a page of results for an API. So I'd call it `GetFlightsResult` or `GetFlightsPage`.

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
```diff
@@ -49,6 +50,9 @@ public FlightFilter() {
    * @return {@code this}, for fluent style
    */
   public FlightFilter addFilterSubmitTime(FlightFilterOp op, Instant timestamp) {
+    if (timestamp == null) {
+      throw new FlightFilterException("Submit filter timestamp cannot be null");
```
pshapiro4broad, line 54: Unless you need a custom exception here, you could use `Objects.requireNonNull()`.

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
```diff
@@ -117,6 +144,9 @@ public FlightFilter addFilterInputParameter(String key, FlightFilterOp op, Objec
     if (key == null) {
       throw new FlightFilterException("Key must be specified in an input filter");
     }
+    if (value == null) {
```
pshapiro4broad, line 147: These new checks all precede a constructor call. It looks like the calling code is trying to enforce a constraint that the constructed object has. If that's the case, the check should be done inside the constructor, not in the caller.

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
pshapiro4broad: Also personally I wouldn't say "cannot be null". I would say "is required". `null` happens to be the way that this code expresses that something is missing, but if the code was changed to use `Optional` then it would no longer use null.

stairway/src/main/java/bio/terra/stairway/impl/StairwayImpl.java
pshapiro4broad: Also note that in general you don't want to provide javadoc for a method override that already has a javadoc. The base class javadoc will be inherited. You'd only want to override the javadoc if your overridden method is different from the base class version in some way.

stairway/src/main/java/bio/terra/stairway/impl/PageToken.java
```diff
@@ -0,0 +1,48 @@
...
+public class PageToken {
+  public static final String PAGE_TOKEN_VERSION = "v01";
+  private final Instant timestamp;
+
+  public PageToken(Instant timestamp) {
+    this.timestamp = timestamp;
+  }
+
+  public PageToken(String token) {
+    if (!StringUtils.startsWith(token, PAGE_TOKEN_VERSION)) {
```
pshapiro4broad, line 22: I don't get why `StringUtils` is used here. Why not `token.startsWith(PAGE_TOKEN_VERSION)` here and `token.substring(PAGE_TOKEN_VERSION.length())` below?

stairway/src/main/java/bio/terra/stairway/Stairway.java
```diff
@@ -245,15 +245,18 @@ public FlightState getFlightState(String flightId)
           InterruptedException;
 
   /**
-   * Enumerate flights - returns a range of flights ordered by submit time. Note that there can be
-   * "jitter" in the paging through flights if new flights are submitted.
+   * NOTE: This endpoint is maintained for backward compatibility. We recommend switching to the new
```
pshapiro4broad, line 248: It's confusing to me that this javadoc is referring to a method as an endpoint. I would expect `endpoint` to be a web API.

stairway/src/main/java/bio/terra/stairway/Stairway.java
```diff
@@ -245,15 +245,18 @@ public FlightState getFlightState(String flightId)
...
+   * getFlights endpoint below.
```
pshapiro4broad, line 249: Do you want to deprecate this API? If so it should have `@Deprecated`  etc etc

stairway/src/main/java/bio/terra/stairway/FlightEnumeration.java
```diff
@@ -0,0 +1,32 @@
...
+
+  public FlightEnumeration(
+      int totalFlights, String nextPageToken, List<FlightState> flightStateList) {
+    this.totalFlights = totalFlights;
+    this.nextPageToken = nextPageToken;
+    this.flightStateList = flightStateList;
+  }
+
+  /** @return total number of flights in the filtered set */
+  public int getTotalFlights() {
```
pshapiro4broad, line 19: Not sure if it's possible/easy but it might be nice to provide information about where the current result is with respect to the end of the pagination. I guess a user could use a null limit to get all the flights after the current page / token, but there's no way to know how many it will get, except for the first paginated result, where they can subtract the number they got back from the total. If you're looking at any other page you don't know how many pages / flights have already been returned.

pshapiro4broad: Looking at this again I think that this isn't really pagination in the way that I was thinking so maybe this doesn't make sense. As it is, the token doesn't encode the time the query was started or the filter used, so there's no guarantee that the next page returned is the same as the next page in the original query.

pshapiro4broad (approved):
I have many comments but all are minor. The feature as it is seems fine to me.

stairway/src/main/java/bio/terra/stairway/impl/StairwayImpl.java
ddietterich: Thanks for this guidance! I removed the duplicate javadoc and added the `@Override` as you suggest below.

stairway/src/main/java/bio/terra/stairway/Stairway.java
ddietterich: I don't have a plan to remove it, so didn't want to deprecate it.

PR #92, opened 2022-01-27
DR-2410 add sort on creation time

nmalfroy:
This PR allows users to specify sort order when enumerating flights where sort if performed on submitted_time (current behavior is to always sort in ascending order).

I put this in the `FlightFilter` class which isn't super obvious but seemed easier than changing the `getFlights` signature.

I also added a (non-unique) index to the flights table on the created_at column since we always sort on it (and at least on TDR prod) the endpoint that calls to that takes over a minute because we quickly get millions of flights

Last thing: I changed up how flight results are compared in the `EnumerateFlightsTest` unit test so that we can test order

### DIFF
--- a/stairway/src/main/java/bio/terra/stairway/FlightFilter.java
+++ b/stairway/src/main/java/bio/terra/stairway/FlightFilter.java
@@ -126,7 +126,7 @@ public class FlightFilter {
    */
   public FlightFilter addFilterFlightStatus(FlightFilterOp op, FlightStatus status) {
     if (status == null) {
-      throw new FlightFilterException("Status cannot be nul");
+      throw new FlightFilterException("Status cannot be null");
     }
     FlightFilterPredicate predicate =
         new FlightFilterPredicate(
@@ -168,7 +168,7 @@ public class FlightFilter {
    */
   public FlightFilter submittedTimeSortDirection(FlightFilterSortDirection submittedTimeSortDirection) {
     if (submittedTimeSortDirection == null) {
-      throw new FlightFilterException("Sort direction cannot be nul");
+      throw new FlightFilterException("Sort direction cannot be null");
     }
 
     this.submittedTimeSortDirection = submittedTimeSortDirection;
--- a/stairway/src/main/java/bio/terra/stairway/FlightFilter.java
+++ b/stairway/src/main/java/bio/terra/stairway/FlightFilter.java
@@ -17,6 +17,7 @@ import javax.annotation.Nullable;
  *     .addFilterFlightClass(EQUAL, IngestFlight.class)
  *     .addFilterFlightStatus(EQUAL, FlightStatus.COMPLETED)
  *     .addFilterInputParameter("email", EQUAL, "ddtest@gmail.com");
+ *     .submittedTimeSortDirection(FlightFilterSortDirection.DESC)
  * }</pre>
  *
  * That filter would return ingest flights completed in the last day (assuming you'd properly set
@@ -26,11 +27,13 @@ import javax.annotation.Nullable;
 public class FlightFilter {
   private final List<FlightFilterPredicate> flightPredicates;
   private final List<FlightFilterPredicate> inputPredicates;
+  private FlightFilterSortDirection submittedTimeSortDirection;
   private int parameterId;
 
   public FlightFilter() {
     flightPredicates = new ArrayList<>();
     inputPredicates = new ArrayList<>();
+    submittedTimeSortDirection = FlightFilterSortDirection.ASC;
     parameterId = 0;
   }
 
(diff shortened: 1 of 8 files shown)


nmalfroy:
Also, I ran:
`./gradlew :stairway:test` locally and it passed. not sure if there are other tests I should run?

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
```diff
@@ -153,6 +160,21 @@ public FlightFilter addFilterInputParameter(String key, FlightFilterOp op, Objec
...
 
+  /**
+   * Specify the sort order based on submitted time for the returned values
+   * @param submittedTimeSortDirection wether to sort in ascending (default) or descending order
+   * @return {@code this}, for fluent style
+   * @throws FlightFilterException if the specified value is null
+   */
+  public FlightFilter submittedTimeSortDirection(FlightFilterSortDirection submittedTimeSortDirection) {
+    if (submittedTimeSortDirection == null) {
+      throw new FlightFilterException("Sort direction cannot be nul");
```
ddietterich: nit: sp `null`

stairway/src/test/java/bio/terra/stairway/impl/EnumerateFlightsTest.java
```diff
@@ -146,14 +148,26 @@ public void enumTest() throws Exception {
...
+                    .submittedTimeSortDirection(FlightFilterSortDirection.ASC);
+    flightList = flightDao.getFlights(0, 100, filter);
+    checkResults("case 11", flightList, Arrays.asList("0", "1", "2", "3", "4", "5"));
+
+    // Case 12: sorting in descending order
+    filter =
+            new FlightFilter()
+                    .submittedTimeSortDirection(FlightFilterSortDirection.DESC);
+    flightList = flightDao.getFlights(0, 100, filter);
+    checkResults("case 12", flightList, Arrays.asList("5", "4", "3", "2", "1", "0"));
```
ddietterich, line 164: Nice!

ddietterich (approved):
LGTM - one typo

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
nmalfroy: doh!  Copy pasta.  I fixed both

jmczerk (approved):
LGTM

stairway/src/test/java/bio/terra/stairway/impl/EnumerateFlightsTest.java
```diff
@@ -146,14 +148,26 @@ public void enumTest() throws Exception {
...
   }
 
   private void checkResults(String name, List<FlightState> resultlList, List<String> expectedIds) {
-    assertThat(
-        name + ": right number of elements", resultlList.size(), equalTo(expectedIds.size()));
     List<String> actualIds =
         resultlList.stream().map(FlightState::getFlightId).collect(Collectors.toList());
-    assertTrue(CollectionUtils.isEqualCollection(actualIds, expectedIds), "elements match");
+    assertThat(name + ": elements match in the correct order", actualIds, contains(expectedIds.toArray()));
```
pshapiro4broad, line 170: Is `.toArray()` needed here?
```suggestion
    assertThat(name + ": elements match in the correct order", actualIds, contains(expectedIds));
```

stairway/src/test/java/bio/terra/stairway/impl/EnumerateFlightsTest.java
```diff
@@ -146,14 +148,26 @@ public void enumTest() throws Exception {
     flightEnum = flightDao.getFlights(pageTokenString, 3, filter);
     checkResults("case 10", flightEnum.getFlightStateList(), Arrays.asList("3", "4", "5"));
     assertThat(flightEnum.getTotalFlights(), equalTo(6));
+
+    // Case 11: sorting in ascending order
+    filter =
+            new FlightFilter()
+                    .submittedTimeSortDirection(FlightFilterSortDirection.ASC);
+    flightList = flightDao.getFlights(0, 100, filter);
+    checkResults("case 11", flightList, Arrays.asList("0", "1", "2", "3", "4", "5"));
```
pshapiro4broad, line 157: You could avoid duplicating the list of flight IDs by using `flights`.
```suggestion
    checkResults("case 11", flightList, flights.stream().map(FlightState::getFlightId).collect(Collectors.toList()));
```

pshapiro4broad (approved):
LGTM, a few minor comments

stairway/src/test/java/bio/terra/stairway/impl/EnumerateFlightsTest.java
nmalfroy: It is...since contains takes a vararg.  If you leave it as a collection, it will treat it as a single item in the list that happens to be a collection (I've been bitten by that many times)

stairway/src/test/java/bio/terra/stairway/impl/EnumerateFlightsTest.java
nmalfroy: Yeah, but I prefer the visibility of seeing what we're testing here

PR #93, opened 2022-02-07
DR-2427 Return Flight className in FlightState object

nmalfroy:
Fairly simple PR that just adds the className to the FlightState object (that data was already stored in the DB)

### DIFF
--- a/stairway/src/test/java/bio/terra/stairway/FlightStateClassTest.java
+++ b/stairway/src/test/java/bio/terra/stairway/FlightStateClassTest.java
@@ -20,10 +20,10 @@ import org.junit.jupiter.api.Test;
 
 @Tag("unit")
 public class FlightStateClassTest {
-  private static String bad = "bad bad bad";
+  private static final String BAD = "bad bad bad";
+  private static final String CLASS_NAME = "bio.terra.stairway.FooFlight";
   private FlightState result;
   private Instant timestamp;
-  private String className;
 
   @BeforeEach
   public void setup() {
@@ -39,8 +39,6 @@ public class FlightStateClassTest {
 
     timestamp = Instant.now();
 
-    className = "bio.terra.stairway.FooFlight";
-
     result = new FlightState();
     result.setFlightId(flightId);
     result.setFlightStatus(FlightStatus.FATAL);
@@ -49,7 +47,7 @@ public class FlightStateClassTest {
     result.setCompleted(timestamp);
     result.setResultMap(outputs);
     result.setException(new RuntimeException(errString));
-    result.setClassName(className);
+    result.setClassName(CLASS_NAME);
   }
 
   @Test
@@ -69,7 +67,7 @@ public class FlightStateClassTest {
     FlightMap outputMap = result.getResultMap().get();
     assertThat(outputMap.get(fkey, Double.class), is(dubValue));
 
-    assertThat(result.getClassName(), is(className));
+    assertThat(result.getClassName(), is(CLASS_NAME));
   }
 
   @Test
@@ -77,7 +75,7 @@ public class FlightStateClassTest {
     Assertions.assertThrows(
         UnsupportedOperationException.class,
         () -> {
-          result.getResultMap().get().put(bad, bad);
+          result.getResultMap().get().put(BAD, BAD);
         });
   }
 
@@ -86,7 +84,7 @@ public class FlightStateClassTest {
     Assertions.assertThrows(
         UnsupportedOperationException.class,
         () -> {
-          result.getInputParameters().put(bad, bad);
+          result.getInputParameters().put(BAD, BAD);
         });
   }
 }
--- a/stairway/src/test/java/bio/terra/stairway/FlightStateClassTest.java
+++ b/stairway/src/test/java/bio/terra/stairway/FlightStateClassTest.java
@@ -23,6 +23,7 @@ public class FlightStateClassTest {
   private static String bad = "bad bad bad";
   private FlightState result;
   private Instant timestamp;
+  private String className;
 
   @BeforeEach
   public void setup() {
@@ -38,6 +39,8 @@ public class FlightStateClassTest {
 
     timestamp = Instant.now();
 
+    className = "bio.terra.stairway.FooFlight";
+
     result = new FlightState();
     result.setFlightId(flightId);
     result.setFlightStatus(FlightStatus.FATAL);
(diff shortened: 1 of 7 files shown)


stairway/src/test/java/bio/terra/stairway/FlightStateClassTest.java
```diff
@@ -38,6 +39,8 @@ public void setup() {
 
     timestamp = Instant.now();
 
+    className = "bio.terra.stairway.FooFlight";
```
ddietterich: This could be `private static final`

ddietterich (approved):
LGTM - one nit

stairway/src/test/java/bio/terra/stairway/FlightStateClassTest.java
nmalfroy: ah good call.  Update that and the `bad` var to be static final

nmalfroy:
Gah, sorry, Dan, didn't mean to re-add you as a reviewer

ddietterich (approved):
Make sure you run spotlessApply before your final merge. Looks like there are some back-and-forth changes that look like spotless

jmczerk:
Just a heads up that I am OOO til Friday, so don't feel a need to wait on my review if you get two approvals before then.

nmalfroy:
Ah good point.  I thought I had but I'll update that.  I'll see about getting another review @jmczerk , thanks!

nmalfroy:
I ran spotless and it looks ok.  I think the back and forth might be from uppercasing the static final

ddietterich:
> 

Thank you

stairway/src/test/java/bio/terra/stairway/impl/EnumerateFlightsTest.java
```diff
@@ -150,24 +148,27 @@ public void enumTest() throws Exception {
...
-                    .submittedTimeSortDirection(FlightFilterSortDirection.ASC);
+    filter = new FlightFilter().submittedTimeSortDirection(FlightFilterSortDirection.ASC);
     flightList = flightDao.getFlights(0, 100, filter);
     checkResults("case 11", flightList, Arrays.asList("0", "1", "2", "3", "4", "5"));
+    // explicitly verify that classnames are returned as expected (note that the flights list is in
+    // ascending order)
+    assertThat(
+        "class names are correct",
+        flightList.stream().map(FlightState::getClassName).collect(Collectors.toList()),
+        contains(flights.stream().map(FlightState::getClassName).toArray()));
```
pshapiro4broad, line 159: It bugs me that there's no better way to do this. It seems like there should be a way to stream both lists at the same time and assertThat() on each pair of elements in order, avoiding the need to create a collection.

stairway/src/test/java/bio/terra/stairway/impl/EnumerateFlightsTest.java
```diff
@@ -150,24 +148,27 @@ public void enumTest() throws Exception {
...
     flightList = flightDao.getFlights(0, 100, filter);
     checkResults("case 12", flightList, Arrays.asList("5", "4", "3", "2", "1", "0"));
   }
 
   private void checkResults(String name, List<FlightState> resultlList, List<String> expectedIds) {
     List<String> actualIds =
         resultlList.stream().map(FlightState::getFlightId).collect(Collectors.toList());
-    assertThat(name + ": elements match in the correct order", actualIds, contains(expectedIds.toArray()));
+    assertThat(
+        name + ": elements match in the correct order", actualIds, contains(expectedIds.toArray()));
```
pshapiro4broad, line 171: It looks like there's a common pattern here that could be abstracted, if you wanted to. The abstraction is the getter on the object, so the general method is something like 
```
assertSameField(List<FlightState> expected, List<FlightState> actual, Function<FlightState, T> getter)
```

stairway/src/test/java/bio/terra/stairway/impl/FilterTest.java
```diff
@@ -214,16 +214,16 @@ public void filterForm1PageTokenTest() throws Exception {
...
-            "SELECT F.flightid, F.stairway_id, F.submit_time, F.completed_time,"
-                    + " F.output_parameters, F.status, F.serialized_exception"
-                    + " FROM flight F"
-                    + " WHERE F.submit_time > :pagetoken"
-                    + " ORDER BY submit_time DESC LIMIT :limit";
+        "SELECT F.flightid, F.stairway_id, F.submit_time, F.completed_time,"
+            + " F.output_parameters, F.status, F.serialized_exception, F.class_name"
+            + " FROM flight F"
+            + " WHERE F.submit_time > :pagetoken"
+            + " ORDER BY submit_time DESC LIMIT :limit";
```
pshapiro4broad, line 221: IMO these kind of string comparison tests aren't very useful and the code is better served by tests that run the SQL. That being said, I would at least create strings for the common substrings to avoid needless code duplication.

pshapiro4broad (approved):
Change looks fine, some comments on the DRY-ness of the test code.

stairway/src/test/java/bio/terra/stairway/impl/FilterTest.java
ddietterich: The main complexity in the filtering code is generating the right SQL. This kind of test makes it much more direct to ensure that is working, rather than debugging SQL errors inside the larger system.

stairway/src/test/java/bio/terra/stairway/impl/EnumerateFlightsTest.java
nmalfroy: I don't know...I guess this doesn't bug me that much...My bigger issue is that I just don't like the JUnit matchers...The assertJ matchers have more explicit list checkers

stairway/src/test/java/bio/terra/stairway/impl/EnumerateFlightsTest.java
nmalfroy: Yeah, maybe...though they are subtly different so it may not be super commonizable

pshapiro4broad: If the comparisons are different, you can write
```
assertSameField(List<FlightState> expected, List<FlightState> actual, 
                             Function<FlightState, T> getter, Matcher<T>)
```

PR #94, opened 2022-05-06
[PF-1534] progress meters

ddietterich:
This PR adds support for recording flight progress.

### Access to Flight Progress in Flight Steps
The FlightContext provides two new methods:
- `void setProgressMeter(String name, long v1, long v2)` - to set or update named progress data. This will immediately write the updated progress to the Stairway database.
- `ProgressMeter getProgressMeter(String name)` - to read progress. The `ProgressMeter` holds `v1` and `v2`.

### Access to Flight Progress from Flight State
The `FlightState` class provides a getter that returns a `ProgressMeterReader` object. That class provides two methods:
- `ProgressMeter getProgressMeter(String name)` - to retrieve named progress data
- `ProgressMeter getFlightStepProgressMeter()` - to retrieve the Stairway flight step progress data

The Stairway flight step data reflects the current step index, rather than the last committed step.

### Implementation Notes
The new interfaces only expose progress. That restriction is driven by my concern that persistent writes that are out-of-band of the saga transaction will encourage bad flight design. However, if other use cases for persistent state show up, I didn't want to rototill Stairway. So I made the underlying implementation use a subclass of FlightMap (`PersistentStateMap`) to hold the progress meters. It would be small work to expose the whole map, if that ever turns out to be useful.

The progress meters are held in a Java `Map<String, ProgressMeterData>`. That is stored in the PersistentStateMap with the key `StairwayProgressMeters`.

The class `ProgressMetersImpl` holds the `PersistentStateMap` and manages the reading, writing, and flushing of that map. It is the class used in `FlightContextImpl` and `FlightState`.


stairway/src/main/java/bio/terra/stairway/impl/FlightContextImpl.java
```diff
@@ -280,8 +309,11 @@ boolean isDoingLastStep() {
     return (isDoing() && getStepIndex() == steps.size() - 1);
   }
 
-  void nextStepIndex() {
+  void nextStepIndex() throws InterruptedException {
     logState.nextStepIndex();
+    // Update the stairway step progress meter to reflect what we step we are working on
+    setProgressMeter(
+        ProgressMetersImpl.STAIRWAY_STEP_PROGRESS, getStepIndex(), getStepClassNames().size());
```
jmczerk: Very much a nit, but I see at least one other place where we infer total number of steps from `stepClassNames.size()`.  Would it be worth wrapping in a `getStepCount()` for readability?

stairway/src/main/java/bio/terra/stairway/impl/FlightDao.java
```diff
@@ -640,6 +640,40 @@ private FlightContextImpl resumeInner(String stairwayId, String flightId)
...
+            + "DO UPDATE SET value = :value2";
+
+    List<FlightInput> inputList = FlightMapUtils.makeFlightInputList(persistedStateMap);
+
+    try (Connection connection = dataSource.getConnection();
+        NamedParameterPreparedStatement statement =
+            new NamedParameterPreparedStatement(connection, sqlUpsert)) {
+      startTransaction(connection);
+      statement.setString("flightId", flightId);
+
```
jmczerk, line 666: I don't know if it's worth a short comment here, but I always forget about the fact that `NamedParameterPreparedStatement` doesn't allow reuse of named parameter.  This confused me when looking at the `ON CONFLICT...` logic, because I initially thought that `value1` and `value2` had some kind of mapping to `ProgressMeter` `v1/v2` semantics.

jmczerk (approved):

stairway/src/main/java/bio/terra/stairway/FlightContext.java
```diff
@@ -48,4 +49,23 @@ public interface FlightContext {
...
   /** @return pretty string describing the flight */
   String flightDesc();
+
+  /**
+   * Return meter data given the meter name
+   *
+   * @param name name of the meter to lookup
+   * @return optional progress meter data
+   */
+  Optional<ProgressMeterData> getProgressMeter(String name);
```
pshapiro4broad: What's the benefit of returning `Optional` here? The only case where returning `Optional` is needed is if you need to distinguish between a `null` value and a missing value. Is there a case where a `null` result is different from a missing result?

If a caller wants to use `Optional` to examine this result, they can always use `Optional.ofNullable(getProgressMeter(...))`.

stairway/src/main/java/bio/terra/stairway/impl/ProgressMetersImpl.java
```diff
@@ -0,0 +1,67 @@
...
+
+  @Override
+  public Optional<ProgressMeterData> getFlightStepProgressMeter() {
+    return getProgressMeter(STAIRWAY_STEP_PROGRESS);
+  }
+
+  private Map<String, ProgressMeterData> getMetersFromMap() {
+    Map<String, ProgressMeterData> meters =
+        persistedStateMap.get(STAIRWAY_PROGRESS_METERS, new TypeReference<>() {});
+    return Optional.ofNullable(meters).orElseGet(HashMap::new);
```
pshapiro4broad: Another approach would be to add another java `Map` API to `FlightMap`; using `Map.computeIfAbsent()` does the same thing and is more idiomatic for a `Map` API.
```suggestion
    Map<String, ProgressMeterData> meters =
        persistedStateMap.computeIfAbsent(STAIRWAY_PROGRESS_METERS, new TypeReference<>() {}, k -> new HashMap<>());
    return meters;
```

stairway/src/main/java/bio/terra/stairway/FlightContext.java
ddietterich: Hmm... that is not my understanding of the only case for returning Optional. I'm still titrating on its use. I thought that returning Optional was preferred to returning null since it forces the caller to code for a missing return.

But I don't feel strongly about it. I ain't afraid of no nulls!

stairway/src/main/java/bio/terra/stairway/impl/ProgressMetersImpl.java
ddietterich: Or we could have a variant of `get` that returns `Optional<T>` and you could write:
```
meters = persistedStateMap.getOpt(STAIRWAY_PROGRESS_METERS, new TypeReference<>(){}).orElseGet(Hash::new);
```
I think you might have suggested that when we visited the `get` and `getRaw` methods.

stairway/src/main/java/bio/terra/stairway/FlightContext.java
ddietterich: @jmczerk do you have an opinion?

It would probably be more consistent with the rest of the interface to remove the Optional.

stairway/src/main/java/bio/terra/stairway/FlightContext.java
```diff
@@ -48,4 +49,23 @@ public interface FlightContext {
...
+
+  /**
+   * Set a progress meter for this flight. Typically, this is used in the form v1 operations are
+   * complete out of v2 total operations.
```
jaycarlton, line 62: Why not name these `numComplete` and `totalOperations`? I can't think of a legitimate way to use them for something else. If it's just `v1` and `v2` I'm worried we could have two conventions if someone gets them switched.

stairway/src/main/java/bio/terra/stairway/FlightContext.java
jaycarlton: Keep the `Optional`. I never want to see a method return `null` if it can be helped.

stairway/src/main/java/bio/terra/stairway/ProgressMeterData.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * POJO that represents the data for one project meter. A project meter is named and has two data
+ * values. The values are typically used to represent progress in the form of: N things done out of
+ * M total things.
+ */
+public class ProgressMeterData {
+  private final long v1;
+  private final long v2;
+
```
jaycarlton: What do you think about adding a `String message` to this structure? I'd like to be able to indicate the resource ID of the particular resource I'm cloning, for example.

stairway/src/main/java/bio/terra/stairway/impl/FlightContextImpl.java
```diff
@@ -246,6 +257,24 @@ public String flightDesc() {
...
 
+  @Override
+  public Optional<ProgressMeterData> getProgressMeter(String name) {
+    return progressMeters.getProgressMeter(name);
+  }
+
+  /**
+   * Set a progress meter and persist it in the Stairway database
+   *
+   * @param name name of the meter
```
jaycarlton, line 267: So the name only needs to be unique to the particular flight? I.e. I can call all my meters "progress"?

stairway/src/main/java/bio/terra/stairway/impl/FlightDao.java
```diff
@@ -640,6 +640,40 @@ private FlightContextImpl resumeInner(String stairwayId, String flightId)
...
+      for (FlightInput input : inputList) {
+        statement.setString("key", input.getKey());
+        statement.setString("value1", input.getValue());
```
jaycarlton, line 672: why are both value1 and value2 set to the same value?

jaycarlton (approved):

stairway/src/main/java/bio/terra/stairway/FlightContext.java
jmczerk: No strong opinion here.  If you expect people to generally ask for progress meters that exist, I'd say keep it bare in favor of consistency with the rest of the interface.  If you expect consuming code to be asking for meters speculatively, then maybe `Optional` would make more sense?

pshapiro4broad: > Keep the Optional. I never want to see a method return null if it can be helped.

Can you provide a reasoning behind this opinion? My original question still remains: what is the benefit of returning `Optional` over `null`? If the answer is "I don't like `null`", that's not very compelling, and is not aligned with with Java implementor's intention of this language feature.

stairway/src/main/java/bio/terra/stairway/ProgressMeterData.java
```diff
@@ -0,0 +1,28 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * POJO that represents the data for one project meter. A project meter is named and has two data
+ * values. The values are typically used to represent progress in the form of: N things done out of
+ * M total things.
+ */
+public class ProgressMeterData {
```
pshapiro4broad: What's the distinction between `ProgressMeter` and `ProgressMeterData`? If there's none, it would be less wordy and just as clear to call this `ProgressMeter`. Changing this name would make it align with the methods that return it, `getProgressMeter()` and `getFlightStepProgressMeter()`.

stairway/src/main/java/bio/terra/stairway/impl/FlightContextImpl.java
```diff
@@ -89,6 +92,10 @@ public class FlightContextImpl implements FlightContext {
   // Persisted state that changes each time a step is completed and its log record is written.
   private final FlightContextLogState logState;
 
+  // -- persisted state --
+  // Progress Meters
+  private final ProgressMetersImpl progressMeters;
```
pshapiro4broad, line 96: Why is `ProgressMeterImpl` needed here (vs `ProgressReader`)? It looks like the only reason is to expose `setProgressMeter()`, but this could also be part of the `ProgressReader` interface as it doesn't expose an API a caller wouldn't already have access to via `FlightContext`.

stairway/src/main/java/bio/terra/stairway/impl/ProgressMetersImpl.java
```diff
@@ -0,0 +1,67 @@
...
+import bio.terra.stairway.ProgressMeterData;
+import bio.terra.stairway.ProgressMeterReader;
+import com.fasterxml.jackson.core.type.TypeReference;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+public class ProgressMetersImpl implements ProgressMeterReader {
+  /** FlightMap key for persistedStateMap for storing the progress meters */
+  public static final String STAIRWAY_PROGRESS_METERS = "StairwayProgressMeters";
```
pshapiro4broad, line 15: Does stairway reserve keys that begin with a specific prefix? If so, it should check for this prefix to prevent a caller from overwriting a property that stairway "owns".

stairway/src/main/java/bio/terra/stairway/impl/ProgressMetersImpl.java
pshapiro4broad: It's true, adding `Optional` would allow users to distinguish between the case where they stored `null` or didn't store anything at all, where before they had to use `containsKey()`. If we think storing `null` in this way is a common use case, then this would be helpful. From the TDR code's use of stairway, this doesn't seem to be the case -- `null` is interchangeable with an empty value.

The way I look at this is, what is the use case we're trying to address? If it's the ability to conditionally create an object when there's no value currently stored, `computeIfAbsent()` seems like the most direct way to express that. ~Also the semantics of the operations are different. `computeIfAbsent` not only computes the new value, but stores its value in the map. If we had `computeIfAbsent()`, we wouldn't need to call `putMetersToMap()`.~ Looking at this again, I think this isn't true after all: the map needs to be re-written every time its contents is changed, since we persist immediately instead of persisting objects at the end. With those semantics, `computeIfAbsent` doesn't work.

stairway/src/main/java/bio/terra/stairway/impl/PersistedStateMap.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * The persisted state class holds a FlightMap that is written back to the database when it is
+ * explicitly flushed. The persisted state is connected to a flight, but is written independently of
+ * step logging.
+ *
+ * <p>Right now, it is only used within ProgressMetersImpl to support flights recording their
+ * progress. However, if other use cases arise, we may expose it directly to flights.
+ */
+public class PersistedStateMap extends FlightMap {
+  private static final Logger logger = LoggerFactory.getLogger(PersistedStateMap.class);
```
pshapiro4broad: This field is unused in the class and can be removed.
```suggestion
```

stairway/src/main/java/bio/terra/stairway/FlightContext.java
jaycarlton: Before there was Optional, callers had to remember or guess or learn from experience when a function could return null, and implement ugly logic to handle it. Now we have a self documenting, functional construct that supports transformations and explicit presence and absence checks. I think Enterprise Java 3rd edition had a good chapter on it. It's a higher level of abstraction than null too, which is a good thing. I think it's also fair to say we've reduced the occurrence of NPEs in WSM significantly by using it brutally.

pshapiro4broad: > Before there was Optional, callers had to remember or guess or learn from experience when a function could return null

First, `Optional` itself can be `null` -- just by declaring that a function returns `Optional` doesn't mean that someone couldn't return a null value from it. In this case we are writing code for our own consumption, so I think we can do better than guesswork and experimentation to determine how an API works. I would recommend good documentation, and a convention of never returning `null` from a method that returns a collection.

> and implement ugly logic to handle it.

Ugliness is a subjective measure. Functional forms are nice but don't always make for (what I would consider to be) idiomatic Java. Writing Java that's trying to be Scala looks ugly to me.

> It's a higher level of abstraction than null too, which is a good thing.

I disagree with this; I would characterize is as an unnecessary wrapper around objects which, except for the case where one needs to distinguish missing from `null`, adds no value. It sounds like we've had different experiences with coding however:

> I think it's also fair to say we've reduced the occurrence of NPEs in WSM significantly by using it brutally.

Anecdotal data is hard to prove or disprove; all I can say is personally I've worked without using `Optional` for a long time and haven't seen a need for this feature. I've also worked in languages that protect against nulls, like Scala, and have had NPE issues. (I've also used languages that natively support non-nullable types, those are great but Java doesn't have this.)

If we were working in a language that supported non-null types (like C#, Kotlin, etc) or idiomatic null handling (like Scala) then this would make sense. But this is Java and (again IMO) such forms are not idiomatic to Java.

stairway/src/main/java/bio/terra/stairway/impl/FlightContextImpl.java
ddietterich: I did that. There were many places using `size()` either on the step or stepClassName list.

stairway/src/main/java/bio/terra/stairway/FlightContext.java
ddietterich: I removed the Optionals. It is more consistent with the bulk of the Stairway API.

stairway/src/main/java/bio/terra/stairway/impl/ProgressMetersImpl.java
ddietterich: I'm going to leave this as is.

stairway/src/main/java/bio/terra/stairway/FlightContext.java
ddietterich: I am very bad at predicting future usage. I'd rather have `v1` and `v2` and leave it to the caller to ascribe semantics, then have someone using `numComplete` to store a value that does not mean `numComplete`.

stairway/src/main/java/bio/terra/stairway/ProgressMeterData.java
ddietterich: Can you expand a bit on how you would use the message?

stairway/src/main/java/bio/terra/stairway/impl/FlightContextImpl.java
ddietterich: Right. If you only ever have one meter, you can always call it `progress`.

stairway/src/main/java/bio/terra/stairway/impl/FlightDao.java
ddietterich: I added a comment to explain. `NamedParameterPreparedStatement` is not the Spring class, but one private to Stairway that is not as smart.

stairway/src/main/java/bio/terra/stairway/ProgressMeterData.java
ddietterich: Good point. I had split things differently and then redid them. I renamed it to ProgressMeter.

stairway/src/main/java/bio/terra/stairway/impl/FlightContextImpl.java
ddietterich: The code reading the FlightState does not have access to the FlightContext and should not be able to write it.

stairway/src/main/java/bio/terra/stairway/impl/ProgressMetersImpl.java
ddietterich: Good idea. I'll implement that.

stairway/src/main/java/bio/terra/stairway/ProgressMeterData.java
jaycarlton: During a workspace clone, we clone individual resources. I could use the message to hold resource IDs or names that the UI could use to display resource name & stats in a progress dialog. Or I could just put the name directly in the message, or serialize a JSON structure and put it in there.

stairway/src/main/java/bio/terra/stairway/impl/FlightContextImpl.java
pshapiro4broad: Gotcha, that's why it's Reader and not Writer :)

jmczerk (approved):

jaycarlton (approved):

PR #100, opened 2022-09-24
[PF-1540] Downgrade Minor version of google-libraries-bom

dexamundsen:
Downgrade Minor version of google-libraries-bom from 26.1.x to 26.0.*

Reason - 26.1.* includes google.cloud.bigquery / google.cloud.storage dependency versions that transitively include google-api-client version 2.x which is incompatible with AI notebook libraries in WSM, etc
The incompatible tranistive dependencie cannot be however excluded via build script config since a different version of the same dependency is included in the same build script, thus making the exclusion a no-op

This change is needed on TCL as well - https://github.com/DataBiosphere/terra-common-lib/pull/90


ddietterich (approved):
This looks fine. Why was the downgrade needed?

zloery (approved):

dexamundsen:
> 

First - AI Notebook implementations I have come across so far use google.api.client 1.x, and error out f 2.0 (the latest) is included. 

sample error message during runs -

Caused by: java.lang.IllegalStateException: You are currently running with version 2.0.0 of google-api-client. You need at least version 1.31.1 of google-api-client to run version 1.32.1 of the Notebooks API library.

this error is consistently seen in all services / clients that use AI notebook and include the above client. 

Second - Reason why we need this changed all thw way upstream - library.bom 26.1.* includes google.cloud.bigquery / google.cloud.storage dependency versions that transitively include google-api-client version 2.x which is incompatible with AI notebook libraries in WSM, etc
The incompatible tranistive dependencie cannot be however excluded via build script config since a different version of the same dependency is included in the same build script, thus making the exclusion a no-op

example in WSM.integration

implementation tcl:<version wth bom 26.1.x> {
    exclude google.api.client
}
implementation google.api.client<any version included directly or transitively>
-- in the above the exclude statement has NO effect.
hence it is essential to update this in stairway, and then TCL before including TCL in any other

PR #101, opened 2022-10-03
PR-2083 Make sure that we properly sort for the nextPageToken

nmalfroy:
...when sorting in descending submissions time order.

This issue is described in https://broadworkbench.atlassian.net/browse/PF-2083

This was an oversight on my part when I first introduced being able to specify sort order

### DIFF
--- a/stairway/src/test/java/bio/terra/stairway/impl/FilterTest.java
+++ b/stairway/src/test/java/bio/terra/stairway/impl/FilterTest.java
@@ -217,7 +217,7 @@ public class FilterTest {
         "SELECT F.flightid, F.stairway_id, F.submit_time, F.completed_time,"
             + " F.output_parameters, F.status, F.serialized_exception, F.class_name"
             + " FROM flight F"
-            + " WHERE F.submit_time > :pagetoken"
+            + " WHERE F.submit_time < :pagetoken"
             + " ORDER BY submit_time DESC LIMIT :limit";
 
     PageToken pageToken = new PageToken(Instant.now());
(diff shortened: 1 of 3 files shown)


ddietterich (approved):
LGTM - thanks for the test

okotsopoulos (approved):

stairway/src/test/java/bio/terra/stairway/impl/FilterTest.java
```diff
@@ -217,7 +217,7 @@ public void filterForm1OrderTest() throws Exception {
         "SELECT F.flightid, F.stairway_id, F.submit_time, F.completed_time,"
             + " F.output_parameters, F.status, F.serialized_exception, F.class_name"
             + " FROM flight F"
-            + " WHERE F.submit_time > :pagetoken"
+            + " WHERE F.submit_time < :pagetoken"
```
pshapiro4broad, line 220: So this was broken before? If so, it makes me wonder about how much value we get from asserting against string literals.

pshapiro4broad (approved):
Looks OK to me. Would it be too much to test paginated queries both descending and ascending?

nmalfroy:
> Looks OK to me. Would it be too much to test paginated queries both descending and ascending?

Oh, the paginated query already gets tested in the ascending case (case 10) right above the one that I just added

stairway/src/test/java/bio/terra/stairway/impl/FilterTest.java
nmalfroy: Yeah, I think I see the value as a sanity check around the sql that gets built.  It's not enough of a test on its own.  The test that I added in the other class is the right place to catch the error that I'm fixing here.

PR #98, opened 2022-09-19
[PF-1542] stairway: gradle and java version upgrades

dexamundsen:
Below changes are made to keep this consistent with other services

- Gradle upgrade to 7.4.1
- JDK upgrade from 11 to 17 (set toolchain compatibility)
- Gradle build cleanup (based on https://github.com/DataBiosphere/terra-common-lib/pull/26 and other Terra repos)
- Add dep locking
- Address javadoc warnings - https://broadworkbench.atlassian.net/browse/PF-2027

Tests successful on local run --> 

image

single test re-run -->
14 actionable tasks: 1 executed, 13 up-to-date
5:16:30 AM: Execution finished ':stairway:test --tests "bio.terra.stairway.SleepQueueTest.clusterSuccessTest"'.

PR #102, opened 2022-10-31
PF-2173 Add more filtering support

nmalfroy:
This PR introduces some new filtering concepts:
- Adds supports for `IN` operations
- Allows filtering on a list of flights
- Adds the ability to specify arbitrary boolean filters when filtering for flights.  This allows users to construct their own complex filters on input parameters.  E.g.
<pre><code><del>FlightFilter flightFilter =
        new FlightFilter()
            .setFilterInputBooleanOperationParameter(
                f ->
                    createOr(
                        createPredicate(
                            f.makeInputPredicate("foo", FlightFilterOp.EQUAL, null)),
                        createAnd(
                            f.makeInputPredicate("foo", FlightFilterOp.EQUAL, "val"),
                            f.makeInputPredicate("bar", FlightFilterOp.LESS_EQUAL, 42))));
</del>
FlightFilter filter = new FlightFilter(
    makeAnd(
        makePredicateFlightStatus(EQUAL, FlightStatus.ERROR),
        makePredicateInput("other_num", LESS_THAN, 456)),
        makeOr(
            makePredicateInput("email", EQUAL, "ddtest@gmail.com"),
            makePredicateInput("num", GREATER_THAN, 123)))
    .submittedTimeSortDirection(FlightFilterSortDirection.DESC);
</code></pre>
(note the tests have some good examples...happy to add more if needed)
- Improves support for `IS NULL` operations on input parameters (the parameter no longer needs to exist)
- Note: this gets rid of the form1-form2-form3 sql generation in favor or a single approach that chains together `EXISTS` statements.  The PR creates an index on the flight input key field to speed up the approach. I tested on a stairway DB with ~5M flights and ~40M rows in the flight inputs table and the performance is pretty good (~1.5 seconds when doing basic filtering)
- It's worth noting: I left the current mode of adding input filters too for backwards compatibility

### DIFF
--- a/stairway/src/main/java/bio/terra/stairway/FlightFilter.java
+++ b/stairway/src/main/java/bio/terra/stairway/FlightFilter.java
@@ -136,6 +136,24 @@ public class FlightFilter {
   }
 
   /**
+   * Filter by flight ids
+   *
+   * @param flightIds A list of flight ids to filter by
+   * @return {@code this}, for fluent style
+   */
+  public FlightFilter addFilterFlightIds(List<String> flightIds) {
+    if (flightIds == null) {
+      throw new FlightFilterException("flightIds can not be null");
+    }
+    FlightFilterPredicate predicate =
+            new FlightFilterPredicate(
+                    "flightid", FlightFilterOp.IN, flightIds, Datatype.LIST, makeParameterName());
+    flightPredicates.add(predicate);
+    return this;
+  }
+
+
+  /**
    * Filter by an input parameter. This is processed by converting the {@code value} into JSON and
    * doing a string comparison against the input parameter stored in the database. The {@code value}
    * object must be <b>exactly</b> the same class as the input parameter.
@@ -154,8 +172,12 @@ public class FlightFilter {
     if (value == null) {
       throw new FlightFilterException("Value cannot be null in an input filter");
     }
-    FlightFilterPredicate predicate =
-        new FlightFilterPredicate(key, op, value, Datatype.STRING, makeParameterName());
+    FlightFilterPredicate predicate;
+    if (op == FlightFilterOp.IN) {
+      predicate = new FlightFilterPredicate(key, op, value, Datatype.LIST, makeParameterName());
+    } else {
+      predicate = new FlightFilterPredicate(key, op, value, Datatype.STRING, makeParameterName());
+    }
     inputPredicates.add(predicate);
     return this;
   }
@@ -233,6 +255,7 @@ public class FlightFilter {
     public enum Datatype {
       STRING,
       TIMESTAMP,
+      LIST,
       NULL
     }
   }
--- a/stairway/src/main/java/bio/terra/stairway/FlightFilter.java
+++ b/stairway/src/main/java/bio/terra/stairway/FlightFilter.java
@@ -2,6 +2,9 @@ package bio.terra.stairway;
 
 import bio.terra.stairway.FlightFilter.FlightFilterPredicate.Datatype;
 import bio.terra.stairway.exception.FlightFilterException;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
@@ -42,6 +45,9 @@ public class FlightFilter {
   private final List<FlightFilterPredicate> inputPredicates;
   private FlightBooleanOperationExpression inputBooleanOperationExpression;
   private FlightFilterSortDirection submittedTimeSortDirection;
+
+  // Mapper should be used to deserialize to generic Postgres JSON
+  private static final ObjectMapper pgJsonMapper = new ObjectMapper();
 
   public FlightFilter() {
     flightPredicates = new ArrayList<>();
@@ -227,7 +233,38 @@ public class FlightFilter {
     return this;
   }
 
-  public interface FlightFilterPredicateInterface {}
+  public record Value(FlightFilterPredicate predicate, String string, Instant instant) {
+    Value(FlightFilterPredicate predicate, String string) {
+      this(predicate, string, null);
+    }
+
+    Value(FlightFilterPredicate predicate, Instant instant) {
+      this(predicate, null, instant);
+    }
+
+    Value() {
+      this(null, null, null);
+    }
+  }
+
+  public List<Value> getValues() throws JsonProcessingException {
+    List<Value> values = new ArrayList<>();
+    for (FlightFilterPredicate predicate : flightPredicates) {
+      values.addAll(predicate.getValues());
+    }
+    for (FlightFilterPredicate predicate : inputPredicates) {
+      values.addAll(predicate.getValues());
+    }
+
+    if (inputBooleanOperationExpression != null) {
+      values.addAll(inputBooleanOperationExpression.getValues());
+    }
+    return values;
+  }
+
+  public interface FlightFilterPredicateInterface {
+    List<Value> getValues() throws JsonProcessingException;
+  }
 
   /**
    * Predicate comparison constructor
@@ -375,6 +412,44 @@ public class FlightFilter {
       return makePredicateFlight("submit_time", op, timestamp, Datatype.TIMESTAMP);
     }
 
+    @Override
+    public List<Value> getValues() throws JsonProcessingException {
+      return List.of(switch (type) {
+        case INPUT -> getInputValues();
+        case FLIGHT -> getFlightValues();
+      });
+    }
+
+    private Value getInputValues() throws JsonProcessingException {
+      return switch (datatype) {
+      case LIST -> {
+        List<String> values = new ArrayList<>();
+        for (Object obj : ((List<?>) value)) {
+          // Need to double encode in order for strings to match
+          values.add(
+                  StairwayMapper.getObjectMapper()
+                          .writeValueAsString(StairwayMapper.getObjectMapper().writeValueAsString(obj)));
+        }
+        yield new Value(this, "[" + String.join(",", values) + "]");
+      }
+      case NULL -> new Value();
+      default -> new Value(this, StairwayMapper.getObjectMapper().writeValueAsString(value));
+      };
+    }
+
+    /**
+     * Get the parameter value for substitution into the prepared statement.
+     */
+    private Value getFlightValues() throws JsonProcessingException {
+      return switch (datatype) {
+        case STRING -> new Value(this, (String) value);
+        case LIST -> new Value(this, pgJsonMapper.writeValueAsString(value));
+        case TIMESTAMP -> new Value(this, (Instant) value);
+        // Ignore the parameter in the null case
+        case NULL -> new Value();
+      };
+    }
+
     public enum Datatype {
       STRING,
       TIMESTAMP,
@@ -410,6 +485,15 @@ public class FlightFilter {
      */
     public static FlightBooleanOperationExpression makeOr(FlightFilterPredicateInterface... expressions) {
       return new FlightBooleanOperationExpression(Operation.OR, List.of(expressions));
+    }
+
+    @Override
+    public List<Value> getValues() throws JsonProcessingException {
+      List<Value> values = new ArrayList<>();
+      for (FlightFilterPredicateInterface expression : expressions()) {
+        values.addAll(expression.getValues());
+      }
+      return values;
     }
 
     public enum Operation {
--- a/stairway/src/main/java/bio/terra/stairway/FlightFilter.java
+++ b/stairway/src/main/java/bio/terra/stairway/FlightFilter.java
@@ -4,7 +4,6 @@ import bio.terra.stairway.FlightFilter.FlightFilterPredicate.Datatype;
 import bio.terra.stairway.exception.FlightFilterException;
 import java.time.Instant;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.function.Function;
 import javax.annotation.Nullable;
@@ -25,6 +24,18 @@ import javax.annotation.Nullable;
  * That filter would return ingest flights completed in the last day (assuming you'd properly set
  * the value of {@code yesterday}) run by user ddtest (assuming you'd passed the email as an input
  * parameter.
+ *
+ * <p>You can also initialize the filter with more complex boolean filters:
+ *
+ * <pre>{@code
+ * FlightFilter filter = new FlightFilter(
+ *     makeAnd(
+ *         makePredicate("other_num", LESS_THAN, 456)),
+ *         makeOr(
+ *             makePredicate("email", EQUAL, "ddtest@gmail.com"),
+ *             makePredicate("num", GREATER_THAN, 123)))
+ *     .submittedTimeSortDirection(FlightFilterSortDirection.DESC)
+ * }</pre>
  */
 public class FlightFilter {
   private final List<FlightFilterPredicate> flightPredicates;
@@ -38,6 +49,16 @@ public class FlightFilter {
     inputPredicates = new ArrayList<>();
     submittedTimeSortDirection = FlightFilterSortDirection.ASC;
     parameterId = 0;
+  }
+
+  /**
+   * Use this constructor to allow for more complex selection criteria. This takes in expression
+   * builder methods that are created using static methods on this class.
+   */
+  public FlightFilter(
+      FlightBooleanOperationExpression.Builder inputBooleanOperationExpressionBuilder) {
+    this();
+    this.inputBooleanOperationExpression = inputBooleanOperationExpressionBuilder.build(this);
   }
 
   public List<FlightFilterPredicate> getFlightPredicates() {
@@ -172,7 +193,7 @@ public class FlightFilter {
    */
   public FlightFilter addFilterInputParameter(String key, FlightFilterOp op, Object value)
       throws FlightFilterException {
-    inputPredicates.add(makeInputPredicate(key, op, value));
+    inputPredicates.add(makeInputPredicate(key, op, value).build(this));
     return this;
   }
 
@@ -224,7 +245,8 @@ public class FlightFilter {
    * @return A newly created FlightFilterPredicate object
    * @throws FlightFilterException if predicate is not supplied
    */
-  public FlightFilterPredicate makeInputPredicate(String key, FlightFilterOp op, Object value) {
+  public static FlightFilterPredicate.Builder makeInputPredicate(
+      String key, FlightFilterOp op, Object value) {
     if (key == null) {
       throw new FlightFilterException("Key must be specified in an input filter");
     }
@@ -234,12 +256,40 @@ public class FlightFilter {
     }
 
     if (op == FlightFilterOp.IN) {
-      return new FlightFilterPredicate(key, op, value, Datatype.LIST, makeParameterName());
+      return new FlightFilterPredicate.Builder(
+          key, op, value, Datatype.LIST, FlightFilter::makeParameterName);
     } else if (value == null) {
-      return new FlightFilterPredicate(key, op, null, Datatype.NULL, null);
+      return new FlightFilterPredicate.Builder(key, op, null, Datatype.NULL, (f) -> null);
     } else {
-      return new FlightFilterPredicate(key, op, value, Datatype.STRING, makeParameterName());
+      return new FlightFilterPredicate.Builder(
+          key, op, value, Datatype.STRING, FlightFilter::makeParameterName);
     }
+  }
+
+  /**
+   * Provide builders for expressions to be ANDed together
+   *
+   * @param expressions Builders for expressions. These are created with static methods on this
+   *     class
+   * @return A newly created expression builder
+   */
+  public static FlightBooleanOperationExpression.Builder makeAnd(
+      FlightFilterPredicateInterface.Builder... expressions) {
+    return new FlightBooleanOperationExpression.Builder(
+        FlightBooleanOperationExpression.Operation.AND, expressions);
+  }
+
+  /**
+   * Provide builders for expressions to be ANDed together
+   *
+   * @param expressions Builders for expressions. These are created with static methods on this
+   *     class
+   * @return A newly created expression builder
+   */
+  public static FlightBooleanOperationExpression.Builder makeOr(
+      FlightFilterPredicateInterface.Builder... expressions) {
+    return new FlightBooleanOperationExpression.Builder(
+        FlightBooleanOperationExpression.Operation.OR, expressions);
   }
 
   private String makeParameterName() {
@@ -247,7 +297,13 @@ public class FlightFilter {
     return "ff" + parameterId;
   }
 
-  public static class FlightFilterPredicate {
+  public interface FlightFilterPredicateInterface {
+    interface Builder {
+      FlightFilterPredicateInterface build(FlightFilter filter);
+    }
+  }
+
+  public static class FlightFilterPredicate implements FlightFilterPredicateInterface {
     private final FlightFilterOp op;
     private final String key;
     private final Object value;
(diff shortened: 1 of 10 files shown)


stairway/src/main/java/bio/terra/stairway/impl/FlightFilterAccess.java
```diff
@@ -226,17 +200,19 @@ private void makeFlightFilter(StringBuilder sb, String prefix) {
...
+        return makeInputPredicateSql(expression.getBasePredicate());
+      case AND, OR:
+        return "("
+            + expression.getExpressions().stream()
+                .map(this::makeBooleanExpressionsFilters)
+                .collect(Collectors.joining(expression.getOperation().getSql()))
+            + ")";
+      default:
+        throw new FlightFilterException("Unrecognized boolean operation");
     }
```
pshapiro4broad: Use a switch expression here to avoid the need to `throw`.
```suggestion
    return switch (expression.getOperation()) {
      case PREDICATE -> makeInputPredicateSql(expression.getBasePredicate());
      case AND, OR -> "("
            + expression.getExpressions().stream()
                .map(this::makeBooleanExpressionsFilters)
                .collect(Collectors.joining(expression.getOperation().getSql()))
            + ")";
    }
```

stairway/src/main/java/bio/terra/stairway/impl/FlightFilterAccess.java
```diff
@@ -226,17 +200,19 @@ private void makeFlightFilter(StringBuilder sb, String prefix) {
...
+  private String makeBooleanExpressionsFilters(FlightBooleanOperationExpression expression) {
+    switch (expression.getOperation()) {
+      case PREDICATE:
+        return makeInputPredicateSql(expression.getBasePredicate());
+      case AND, OR:
+        return "("
+            + expression.getExpressions().stream()
+                .map(this::makeBooleanExpressionsFilters)
+                .collect(Collectors.joining(expression.getOperation().getSql()))
+            + ")";
```
pshapiro4broad: You can simplify this by using the `joining()` form that supports prefix and suffix.
```suggestion
        return expression.getExpressions().stream()
                .map(this::makeBooleanExpressionsFilters)
                .collect(Collectors.joining(expression.getOperation().getSql(), "(", ")"));
```

stairway/src/main/java/bio/terra/stairway/impl/FlightFilterAccess.java
```diff
@@ -289,19 +278,66 @@ private void storeFlightPredicateValue(
    */
   @VisibleForTesting
   String makeInputPredicateSql(FlightFilterPredicate predicate) {
-    return "(I.key = '"
+    String valuePredicate = "";
+    if (predicate.getDatatype() == Datatype.NULL) {
+      return "(EXISTS (SELECT 0 FROM flightinput I WHERE F.flightid = I.flightid AND I.key = "
```
pshapiro4broad, line 259: If the subselect is the same, is there a way to refactor the string to avoid the code duplication?

nmalfroy: It's not totally the same and pulling out the similar pieces I thought would make the code less readable

stairway/src/main/java/bio/terra/stairway/impl/FlightFilterAccess.java
nmalfroy: Oh, interesting...I didn't realize you could prefix and suffix there too.  cool

stairway/src/main/java/bio/terra/stairway/impl/FlightFilterAccess.java
nmalfroy: This isn't on java 17 yet so that syntax doesn't work sadly

stairway/src/main/java/bio/terra/stairway/impl/FlightFilterAccess.java
```diff
@@ -289,19 +278,66 @@ private void storeFlightPredicateValue(
...
+    String jsonValue;
+    if (predicate.getDatatype() == Datatype.LIST) {
+      List<String> values = new ArrayList<>();
+      for (Object value : ((List<?>) predicate.getValue())) {
+        // Need to double encode in order for strings to match
+        values.add(
+            StairwayMapper.getObjectMapper()
+                .writeValueAsString(StairwayMapper.getObjectMapper().writeValueAsString(value)));
+      }
+      jsonValue = "[" + StringUtils.join(values, ",") + "]";
```
pshapiro4broad: `StringUtils` is unnecessary here
```suggestion
      jsonValue = "[" + String.join(",", values) + "]";
```

stairway/src/main/java/bio/terra/stairway/impl/FlightFilterAccess.java
pshapiro4broad: Oh wait, is stairway not on JDK 17 yet?

nmalfroy: Yeah.  Libraries get converted later since you don't want to break dependencies and I don't think that all services that use it are on 17 yet

stairway/src/test/java/bio/terra/stairway/impl/EnumerateFlightsTest.java
```diff
@@ -174,6 +176,82 @@ public void enumTest() throws Exception {
...
+
+    flightList = flightDao.getFlights(0, 100, filter);
+    checkResults("case 20", flightList, Arrays.asList("0", "1", "2", "3", "4", "5"));
+
+    // Case 21: filter input on an in clause with a POJO
+    filter =
+        new FlightFilter()
+            .addFilterInputParameter("in2", FlightFilterOp.IN, Arrays.asList(pojo1, pojo2));
+    flightList = flightDao.getFlights(0, 100, filter);
+    checkResults("case 21", flightList, Arrays.asList("1", "2", "3", "4"));
```
pshapiro4broad: It looks like this code supports using boolean operations and other operations (like `addFilterInputParameter()`)  in the same filter. If so, is this functionality tested? I didn't see a test for it here.

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
```diff
@@ -177,6 +213,35 @@ public FlightFilter submittedTimeSortDirection(
...
+      throw new FlightFilterException("Key must be specified in an input filter");
+    }
+    if (value == null && op != FlightFilterOp.EQUAL) {
+      throw new FlightFilterException(
+          "Value cannot be null in an input filter if not doing an equality check");
+    }
+
+    if (op == FlightFilterOp.IN) {
+      return new FlightFilterPredicate(key, op, value, Datatype.LIST, makeParameterName());
+    } else if (value == null) {
```
pshapiro4broad: I'm wondering why you chose to do this, rather than an explicit op like `FlightFilterOp.IS_NULL`. Also, is there a way to check `NOT NULL`?

stairway/src/test/java/bio/terra/stairway/impl/EnumerateFlightsTest.java
nmalfroy: Good eye.  I'll add a test for that

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
nmalfroy: There was already logic for NULL datatypes so it seemed more appropriate to keep that

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
```diff
@@ -233,7 +298,81 @@ public String getParameterName() {
...
+              .toArray(FlightBooleanOperationExpression[]::new));
+    }
+
+    public static FlightBooleanOperationExpression createOr(
+        FlightBooleanOperationExpression... expressions) {
+      return new FlightBooleanOperationExpression(Operation.OR, null, expressions);
+    }
+
+    public enum Operation {
+      PREDICATE(null),
```
pshapiro4broad: It seems like you could support `NOT` by adding `PREDICATE_NOT` here.. not sure how hard it would be to do though, or how useful it would be.

nmalfroy: 100% I was debating doing it but the PR was getting kinda big already so I held off.  If folks think it would be useful, I'm happy to add it too (it's pretty easy)

stairway/src/main/java/bio/terra/stairway/impl/FlightFilterAccess.java
pshapiro4broad: Actually stairway _is_ on JDK 17 as of https://github.com/DataBiosphere/stairway/pull/98

nmalfroy: Oh weird...it was giving me sass

ddietterich (changes requested):
I don't think the is the right filter interface. It would be better if there were one class hierarchy of boolean ops and field predicates that could be constructed by the caller. I want you to be able to say:
```
FlightFilter flightFilter =
  new FlightFilter()
    .makeOr(
       makeInputPredicate("foo", FlightFilterOp.EQUAL, null))
    .makeAnd(
       makeInputPredicate("foo", FlightFilterOp.EQUAL, "val"),
       makeInputPredicate("bar", FlightFilterOp.LESS_EQUAL, 42)));
```
I suggest making a varargs version of the constructor, so that we could write the expression without the dots (e.g., `.makeOr`). So the current constructor would stay as is and take the flat fluent `.make` form and the new constructor would be something like:
```
public FlightFilter(PredicateBaseClass... predicates)
```
and take allow constructing the predicate tree.

nmalfroy:
Yeah, the problem comes with nesting (like how do you nest an or clause within an and clause?  Which is a thing that TDR needs).  We basically need a thing that encompassed the current input predicates and the boolean operator

nmalfroy:
I don't think you could express the filter in the PR description with what you're proposing

ddietterich:
You have to have a base class that can be either a filter or a boolean. Then the (...) lists can take either, like this:
```
FlightFilter flightFilter =
  new FlightFilter(
     makeAnd(
       makeInputPredicate("foo", FlightFilterOp.EQUAL, null)),
       makeInputPredicate("fribble", FlightFilterOp.EQUAL, "Friendly"),
       makeOr(
         makeInputPredicate("foo", FlightFilterOp.EQUAL, "val"),
         makeInputPredicate("bar", FlightFilterOp.LESS_EQUAL, 42),
         makeAnd(
            makeInputPredicate("blah", FlightFilterOp.EQUAL, "Unfriendly"),
            makeInputPredicate("haha", FlightFilterOp.NOT_EQUAL, "funny"))));
```

nmalfroy:
I suppose we could so something like a nested builder model where there's a builder context.  So it could look something like:
```
FlightFilter flightFilter =
  new FlightFilter()
    .startOr()
       .addInputPredicate("foo", FlightFilterOp.EQUAL, null))
       .startAnd()
           .addInputPredicate("foo", FlightFilterOp.EQUAL, "val"),
           .addInputPredicate("bar", FlightFilterOp.LESS_EQUAL, 42)));
       .endAnd()
    .endOr()
```

That seems a little more unwieldy to me but it could work. What do you think?

ddietterich:
You don't like holding the list in an alternate FlightFilter constructor for some reason?

nmalfroy:
> You have to have a base class that can be either a filter or a boolean. Then the (...) lists can take either, like this:
> 
> ```
> FlightFilter flightFilter =
>   new FlightFilter(
>      makeAnd(
>        makeInputPredicate("foo", FlightFilterOp.EQUAL, null)),
>        makeInputPredicate("fribble", FlightFilterOp.EQUAL, "Friendly"),
>        makeOr(
>          makeInputPredicate("foo", FlightFilterOp.EQUAL, "val"),
>          makeInputPredicate("bar", FlightFilterOp.LESS_EQUAL, 42),
>          makeAnd(
>             makeInputPredicate("blah", FlightFilterOp.EQUAL, "Unfriendly"),
>             makeInputPredicate("haha", FlightFilterOp.NOT_EQUAL, "funny"))));
> ```

I can certainly do that but that would be a breaking change and I was trying to avoid that

nmalfroy:
A

> You don't like holding the list in an alternate FlightFilter constructor for some reason?

Oh, I don't mind that, I just was worried having a new base class which would be a breaking change. If we're ok with that, I can totally make a base class that encompasses the boolean operation and the input filter and then allow nesting that way

ddietterich:
Why would it be a breaking change?

I am imagining that the old constructor creates a top-level AND predicate and accepts the old style `.make` methods. The new style can set some state that would disallow the old style methods and only take the nested form.

nmalfroy:
Ah that's an interesting point!  Ok, lemme play with this and I'll update the PR.  Thanks for the suggestion!  I think this'll look a lot cleaner

nmalfroy:
Sorry, getting back to this today.  I'm actually not sure that this is going to work.  In the example that you've for makeAnd, makeOr, makePredicate are all being called from a static context (at least that's what it looks like).  The problem is that an input predicate needs to call the makeParameter method which is tied to an FlightFilter instance...not really sure how to work around short of using a provider pattern but then we're kind right back where we started

nmalfroy:
Actually, I think I got a thing working with a builder pattern.  I'll post an update in a few

nmalfroy:
Ok, I updated the pattern.  @ddietterich, curious what you think now?

nmalfroy:
@ddietterich Just sending a ping to see if you've had a chance to re-review

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
```diff
@@ -148,15 +193,27 @@ public FlightFilter addFilterFlightStatus(FlightFilterOp op, FlightStatus status
...
+   * Filter by a boolean operation. This is an OR or AND statement with a list of
+   * FlightBooleanOperationExpression objects or a PREDICATE that just contains a single
+   * FlightFilterPredicate (it's a wrapper to allow to nest general FlightFilterPredicate objects in
+   * these boolean operations).
+   *
+   * @param expression The boolean operation expression to add
+   * @return {@code this}, for fluent style
+   * @throws FlightFilterException if predicate is not supplied
+   */
+  public FlightFilter setFilterInputBooleanOperationParameter(
```
ddietterich: InterlliJ thinks this call is unused. Delete?

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
```diff
@@ -23,10 +24,23 @@
...
  * parameter.
+ *
+ * <p>You can also initialize the filter with more complex boolean filters:
+ *
+ * <pre>{@code
+ * FlightFilter filter = new FlightFilter(
+ *     makeAnd(
+ *         makePredicate("other_num", LESS_THAN, 456)),
+ *         makeOr(
+ *             makePredicate("email", EQUAL, "ddtest@gmail.com"),
```
ddietterich: The method is actually, `makeInputPredicate`. That raises the question, how do flight predicates fit into this?
I was expecting to see a `makeFlightPredicate` method to match the `makeInputPredicate` method that would integrate the flight predicates into the general boolean expression.

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
```diff
@@ -177,12 +234,76 @@ public FlightFilter submittedTimeSortDirection(
...
+        FlightBooleanOperationExpression.Operation.OR, expressions);
+  }
+
   private String makeParameterName() {
     parameterId++;
     return "ff" + parameterId;
   }
 
-  public static class FlightFilterPredicate {
+  public interface FlightFilterPredicateInterface {
```
ddietterich: IIUC the only reason for the Builder model is to do name assignments. Seems like we could do that during filter SQL generation instead. It just happened to be easier to do it in the FlightFilter class using the original technique. It makes this code rather clunky.

stairway/src/main/java/bio/terra/stairway/impl/FlightFilterAccess.java
```diff
@@ -150,52 +155,22 @@ String makeSql() {
...
-    sb.append(FlightDao.FLIGHT_TABLE)
-        .append(" F INNER JOIN (SELECT flightid, COUNT(*) AS matchCount")
-        .append(" FROM ")
-        .append(FlightDao.FLIGHT_INPUT_TABLE)
-        .append(" I WHERE ");
-    makeInputFilter(sb);
-    sb.append(" GROUP BY I.flightid) INPUT ON F.flightid = INPUT.flightid")
-        .append(" WHERE INPUT.matchCount = ")
-        .append(filter.getInputPredicates().size());
     makeFlightFilter(sb, " AND ");
```
ddietterich, line 174: This means the flight filters cannot participate in the boolean structure. Only added in as a top-level AND.

How about this idea: instead of segregating the generation by maintaining two separate arrays, maybe remember the source of the data in the `FlightFilterPredicate`. We could have the` FlightFilterPredicate` take a `keyType` enum argument. Key type would be FLIGHT_TABLE or INPUT_PARAMETER.

In the future, we could add filtering by the output params of completed flights by adding an OUTPUT_PARAMETER type.

The current `add` method results could be collected into a single list of `List<FlightFilterPredicateInterface.Builder>`.

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
```diff
@@ -233,7 +354,96 @@ public String getParameterName() {
...
+      }
+    }
+  }
+
+  public static class FlightBooleanOperationExpression implements FlightFilterPredicateInterface {
+
+    private final Operation operation;
+    private final List<FlightFilterPredicateInterface> expressions;
+
+    private FlightBooleanOperationExpression(
```
ddietterich: IntelliJ also thinks this is never used.

stairway/src/main/java/bio/terra/stairway/impl/FlightFilterAccess.java
```diff
@@ -70,50 +83,42 @@ void storePredicateValues(NamedParameterPreparedStatement statement)
...
+   * <p>Input parameter filters are ANDed together by default or can be organized using
+   * FlightBooleanOperationExpression objects. The generated SQL might look like
    *
    * <pre>{@code
-   * SELECT flight.*
-   * FROM flight JOIN flight_input
-   *   ON flight.flightid = flight_input.flightid
-   * WHERE flight-filters
-   *   AND flight_input.key = 'key' AND flight_input.value = 'json-of-object'
+   * SELECT *
```
ddietterich, line 89: These examples only cover the cases where the predicates are on input values.
They need to cover the cases where the predicates are on flight columns.

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
```diff
@@ -233,7 +354,96 @@ public String getParameterName() {
     public enum Datatype {
       STRING,
       TIMESTAMP,
+      LIST,
```
ddietterich, line 456: What this type means isn't obvious to me. And the semantics in FlightFilterAccess that convert to JSON are also not obvious. Maybe some documentation about what this means?

I'm noticing that documentation about using the FlightFilter is only in the comment at the top of this module and is not comprehensive. Not your problem, but something for me to add to the Stairway list.

stairway/src/main/java/bio/terra/stairway/impl/FlightFilterAccess.java
```diff
@@ -289,19 +277,70 @@ private void storeFlightPredicateValue(
...
+          + "' AND I.value IS NULL) "
+          + "OR NOT EXISTS (SELECT 0 FROM flightinput I WHERE F.flightid = I.flightid AND I.key = "
+          + "'"
+          + predicate.getKey()
+          + "'))";
+    } else if (predicate.getDatatype() == Datatype.LIST) {
+      valuePredicate =
+          "I.value "
+              + predicate.getOp().getSql()
+              + " (SELECT JSON_ARRAY_ELEMENTS_TEXT(CAST(:"
```
ddietterich, line 271: This is pretty odd. What is this supposed to do?

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
```diff
@@ -233,7 +354,96 @@ public String getParameterName() {
...
       NULL
     }
+
+    public static class Builder implements FlightFilterPredicateInterface.Builder {
```
ddietterich: Everything in here is public. I wonder if it needs to be.

ddietterich (changes requested):
At long last, I got to look at this. I like this approach. Now my concerns are mostly at the implementation level.

My primary concern is the lack of support for flight column predicates in the boolean form. I think there is a nice approach to addressing this, outlined in comments.

I also think the Builder model is not needed and can be accommodated in the FlightFilterAccess code. That would be a nice simplification.

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
```diff
@@ -177,12 +236,180 @@ public FlightFilter submittedTimeSortDirection(
...
+  }
+
   private String makeParameterName() {
     parameterId++;
     return "ff" + parameterId;
   }
 
-  public static class FlightFilterPredicate {
+  public interface FlightFilterPredicateInterface {
+    interface Builder {
```
pshapiro4broad: Note that class members like `Builder` aren't inherited so there's no reason or benefit to have this as a member, other than needing to type `FlightFilterPredicateInterface.Builder` instead of `Builder` when you refer to it.

It's also suspicious to me that `FlightFilterPredicateInterface` is an interface with no members. Looking at the code, I think it's possible to make `storeInputPredicateValue()` and `storeFlightPredicateValue()` both methods on these objects, vs methods in `FlightFilterAccess`. If that works, it should make the code simpler.

nmalfroy: It's really there to enable backwards compatibility with existing filtering methods.  I want this PR to not be a breaking change for existing implementors and this should allow that

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
pshapiro4broad: I find the current code's use of `builder` to be confusing and doesn't seem to be doing much. The code has general support for  creating a parameter name but in the code base the only implementation of this is `FlightFilter.makeParameterName()`. Could the code instead call this method at the time the predicate is created?

pshapiro4broad: I've refactored the code locally to no longer use `builder` and it seems to work fine (that is, all the stairway tests are still passing). The only downside I can see is that creating a filter like this is slightly more awkward. Instead of
```
    FlightFilter filter =
        new FlightFilter(
            makeOr(
                makePredicateFlightClass(FlightFilterOp.EQUAL, TestFlight.class),
                makePredicateInput("email", FlightFilterOp.EQUAL, "ddtest@gmail.com"),
                makeAnd(
                    makePredicateInput("name", FlightFilterOp.EQUAL, "dd"),
                    makePredicateInput("resource", FlightFilterOp.EQUAL, "resourceId"))));
```

I have
```
    FlightFilter filter = new FlightFilter();
    filter.setInputBooleanOperationExpression(
        filter.makeOr(
            filter.makePredicateFlightClass(FlightFilterOp.EQUAL, TestFlight.class),
            filter.makePredicateInput("email", FlightFilterOp.EQUAL, "ddtest@gmail.com"),
            filter.makeAnd(
                filter.makePredicateInput("name", FlightFilterOp.EQUAL, "dd"),
                filter.makePredicateInput("resource", FlightFilterOp.EQUAL, "resourceId"))));
```

pshapiro4broad: If instead the parameter was generated during sql generation as @ddietterich suggests then the `make` methods could become static again.

pshapiro4broad: I have a PR onto this branch with these changes https://github.com/DataBiosphere/stairway/pull/111

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
pshapiro4broad: I must be missing something, can you provide an example where changing this would break backward compatibility?

stairway/src/main/java/bio/terra/stairway/impl/FlightFilterAccess.java
nmalfroy: oh, that's in order to work with the JSON encoded data in the flight input table

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
nmalfroy: Yes, FlightFilterPredicate is a public interface (not java interface but an object that a user can instantiate). By turning it into an interface, it makes it so that existing code should be able to be left unchanged

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
nmalfroy: I updated the documentation at the top of FlightFilter to address this. I'll add more inline with the code.  Maybe list could be array (which is a little more generic).  I didn't make the LIST type parameterized since I and only imagine it would ever be used with strings (I don't think we need to support List of timestamps for querying)

stairway/src/main/java/bio/terra/stairway/impl/FlightFilterAccess.java
nmalfroy: I'll update the comments here

stairway/src/main/java/bio/terra/stairway/impl/FlightFilterAccess.java
nmalfroy: I updated the PR so that inputs can be included too.  That was a great call to add.  I've update the unit tests to have examples of those being mixed in too

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
nmalfroy: I moved away from a builder so this should be a moot point now

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
nmalfroy: I should amend this: by having `FlightFilterPredicateInterface` this allows nesting or/and clauses by having `FlightBooleanOperationExpression` include a list of `FlightFilterPredicateInterface` (which can be `FlightBooleanOperationExpression` or `FlightFilterPredicate` objects) as it's children

stairway/src/main/java/bio/terra/stairway/impl/FlightFilterAccess.java
nmalfroy: I updated the comments to show more examples

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
nmalfroy: that should be cleaned up now

stairway/src/main/java/bio/terra/stairway/impl/FlightFilterAccess.java
nmalfroy: I had thought about that but then you get a weird case with supporting the existing API that returns either input or flight filters.  I went a separate approach and left those but added a top level boolean expression that now takes input and flight (and could potentially take output) filters.  So it's really two ways to construct filtering

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
nmalfroy: Yep, totally agreed.  I updated so that now predicates can be flight and input predicates

nmalfroy: (I updated the doc string to show that too)

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
nmalfroy: yep. removed

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
```diff
@@ -23,18 +26,42 @@
...
     submittedTimeSortDirection = FlightFilterSortDirection.ASC;
-    parameterId = 0;
+  }
+
+  /**
+   * Use this constructor to allow for more complex selection criteria. This takes in expression
+   * builder methods that are created using static methods on this class.
+   */
+  public FlightFilter(FlightBooleanOperationExpression inputBooleanOperationExpression) {
+    this();
```
pshapiro4broad: In general I think it's better to chain from fewer params to more, so it's explicit that `inputBooleanOperationExpression` is `null` in one case. Then you could make `inputBooleanOperationExpression` `final`.

You could also also change the other fields to be initialized at their declarations, but it could work either way.
```
public class FlightFilter {
  private final List<FlightFilterPredicate> flightPredicates = new ArrayList<>();
  private final List<FlightFilterPredicate> inputPredicates = new ArrayList<>();
  private final FlightBooleanOperationExpression inputBooleanOperationExpression;
  private FlightFilterSortDirection submittedTimeSortDirection = FlightFilterSortDirection.ASC;

  // Mapper should be used to deserialize to generic Postgres JSON
  private static final ObjectMapper pgJsonMapper = new ObjectMapper();

  public FlightFilter() {
    this(null);
  }

  /**
   * Use this constructor to allow for more complex selection criteria. This takes in expression
   * builder methods that are created using static methods on this class.
   */
  public FlightFilter(FlightBooleanOperationExpression inputBooleanOperationExpression) {
    this.inputBooleanOperationExpression = inputBooleanOperationExpression;
  }
```

pshapiro4broad (approved):
The diff got pretty big but I think it's a little clearer this way. One last comment about constructor chaining, otherwise looks OK to me.

stairway/src/main/java/bio/terra/stairway/FlightFilter.java
nmalfroy: ah good call!

ddietterich (approved):
Thank you Nick and Phil!

PR #104, opened 2023-01-25
[PF-2421] Update dependency versions

dexamundsen:
- Update dependency versions
- Standardize build file set up
- Upgrading versions in TCL, WSM, etc will benefit by upgrading them from core libraries here

Ran tests - clusterSuccessTest is flaky and at times fails when run along with others, but succeeds individually. Likely due to the waitTimeout duration

image

image


dexamundsen:
@ddietterich heads up. Testing this in WSM prior to submitting it for review.

dexamundsen:
will be handled downstream after private fork

PR #105, opened 2023-03-15
[PF-2594] fix undo error overwriting original failure

ddietterich:
The flight runner was updating the result with the dismal failure, replacing the initial failure. That was not the intended behavior. The fix is removing one line.

I added a test that verifies the right error is returned in the flight status.


zloery (approved):

nmalfroy (approved):
good catch!

PR #109, opened 2023-04-12
[TERRA-452], [TERRA-485]. [TERRA-492] Add sonarqube, spotless config 

dexamundsen:
sonar/spotless config copied from - https://github.com/DataBiosphere/terra-aws-resource-discovery
added / configured jacoco
minor cleanup / refactor of gradle files

[TERRA-492] Use latest bumper version - fixes github set-output deprecated warnings


dexamundsen:
will be handled downstream after private fork

PR #111, opened 2023-08-11
Boolean query support without a builder pattern

pshapiro4broad:
Using the query support changes as a start, move parameter name creation to the SQL code in `FlightFilterAccess` to avoid needing state to create a boolean filter predicate.

### DIFF
--- a/stairway/src/main/java/bio/terra/stairway/FlightFilter.java
+++ b/stairway/src/main/java/bio/terra/stairway/FlightFilter.java
@@ -5,7 +5,6 @@ import bio.terra.stairway.exception.FlightFilterException;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.function.Function;
 import javax.annotation.Nullable;
 
 /**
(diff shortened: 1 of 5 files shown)


stairway/src/main/java/bio/terra/stairway/FlightFilter.java
pshapiro4broad, line 1: I had many more changes here but I reverted most of them to make the diff clearer. The main difference is to remove builder and the parameter name generation.

I also changed the predicate classes to be records and moved the factory methods for the classes inside the classes they're creating.

stairway/src/test/java/bio/terra/stairway/impl/FilterTest.java
```diff
@@ -38,13 +38,13 @@ public void predicateTest() throws Exception {
   }
 
   private void testPredicate(FlightFilter filter, int index, String operand) {
-    String flightCompareSql = String.format("F.afield %s :ff%d", operand, index + 1);
+    String flightCompareSql = String.format("F.afield %s :ff%d", operand, 1);
```
pshapiro4broad, line 41: With this change the parameter IDs aren't incremented across queries; each instance of `FlightFilterAccess` has its own counter for IDs. It looks like each query is run independently so this should be OK to do. If it isn't, a global parameter counter could be used to ensure that IDs are unique across all queries.

nmalfroy (approved):
This is a great idea.  Thanks for the change!  I'll merge this into the main PR

PR #112, opened 2023-08-15
Further refactor filter code to avoid needing to type-switch

pshapiro4broad:
Further refactor filter code to avoid needing to type-switch explicitly (in storeInputPredicateValue) and implicitly (due to overriding on the first argument of storeInputPredicateValue()).

The code has two types of polymorphic behavior that don’t use OOP, but could. Using OOP makes the polymorphism more apparent. The code has two methods on `FlightFliterAccess`:
```
storeInputPredicateValue(FlightBooleanOperationExpression booleanExpression, NamedParameterPreparedStatement statement)
storeInputPredicateValue(FlightFilterPredicate predicate, NamedParameterPreparedStatement statement)
```
In this case `storeInputPredicateValue` is polymorphic with respect to its first argument. Since both of these objects have a common base class (`FlightFilterPredicateInterface`), if you could move the methods out of `FlightFilterAccess` (and into the common base class) you can use OOP to do the polymorphic dispatch for you. To do that you either have to pass in a lambda to set the parameter, or return an object that can be used to later set the parameter. I tried it both ways and returning the object seemed to work better.

The second polymorphism is inside `storeInputPredicateValue(FlightBooleanOperationExpression booleanExpression, NamedParameterPreparedStatement statement)` where it does a type-switch on the subexpression type. It turned out that this could also be replaced with a method call on the base class’s API.

### DIFF
--- a/stairway/src/main/java/bio/terra/stairway/FlightFilter.java
+++ b/stairway/src/main/java/bio/terra/stairway/FlightFilter.java
@@ -2,6 +2,9 @@ package bio.terra.stairway;
 
 import bio.terra.stairway.FlightFilter.FlightFilterPredicate.Datatype;
 import bio.terra.stairway.exception.FlightFilterException;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


stairway/src/main/java/bio/terra/stairway/FlightFilter.java
```diff
@@ -375,6 +412,44 @@ public static FlightFilterPredicate makePredicateSubmitTime(
...
+      };
+    }
+
+    /**
+     * Get the parameter value for substitution into the prepared statement.
+     */
+    private Value getFlightValues() throws JsonProcessingException {
+      return switch (datatype) {
+        case STRING -> new Value(this, (String) value);
+        case LIST -> new Value(this, pgJsonMapper.writeValueAsString(value));
```
nmalfroy, line 446: Any reason not to use `StairwayMapper.getObjectMapper()` here?

pshapiro4broad: I don't know, my goal was only to refactor so in this case I preserved what the original code was doing. Maybe @ddietterich could say for sure?

nmalfroy: I suppose we should just leave that for now and we can come back to fix if needed

nmalfroy (approved):
Cool, this makes sense.  thanks!

PR #113, opened 2023-11-09
[PF-2976] Add Azure Service Bus support to stairway

giventocode:
This PR:
- Introduces a module `stairway-azure` that implements `QueueInterface` for Azure Service Bus using a topic/subscription model.
- The module allows authentication with either managed identity (MI) or a connection string.
- Read operations use peek-lock semantics, meaning that messages are sent back to the subscription (queue) in case of processing failures.

### DIFF
--- a/stairway-azure/src/main/java/bio/terra/stairway/azure/AzureServiceBusQueue.java
+++ b/stairway-azure/src/main/java/bio/terra/stairway/azure/AzureServiceBusQueue.java
@@ -10,6 +10,7 @@ import com.azure.messaging.servicebus.ServiceBusMessage;
 import com.azure.messaging.servicebus.ServiceBusReceivedMessage;
 import com.azure.messaging.servicebus.ServiceBusReceiverClient;
 import com.azure.messaging.servicebus.ServiceBusSenderClient;
+import com.azure.messaging.servicebus.models.ServiceBusReceiveMode;
 import java.time.Duration;
 import java.util.List;
 import org.apache.commons.lang3.NotImplementedException;
@@ -65,6 +66,8 @@ public class AzureServiceBusQueue implements QueueInterface {
             .receiver()
             .topicName(builder.topicName)
             .subscriptionName(builder.subscriptionName)
+            .receiveMode(ServiceBusReceiveMode.PEEK_LOCK)
+            .disableAutoComplete()
             .maxAutoLockRenewDuration(builder.maxAutoLockRenewDuration)
             .buildClient();
 
--- a/stairway-azure/src/main/java/bio/terra/stairway/azure/AzureServiceBusQueue.java
+++ b/stairway-azure/src/main/java/bio/terra/stairway/azure/AzureServiceBusQueue.java
@@ -35,7 +35,7 @@ public class AzureServiceBusQueue implements QueueInterface {
    * @param serviceBusReceiverClient the receiver client
    * @param serviceBusSenderClient the sender client
    */
-  public AzureServiceBusQueue(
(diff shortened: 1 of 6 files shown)


stairway-azure/src/main/java/bio/terra/stairway/azure/AzureServiceBusQueue.java
```diff
@@ -0,0 +1,240 @@
...
+      boolean processSucceeded = processFunction.apply(message.getBody().toString());
+
+      if (processSucceeded) {
+        serviceBusReceiverClient.complete(message);
+        logger.info("Completed message. ID: {}", message.getMessageId());
+        return;
+      }
+
+      logger.info("Failed to process message. ID: {}", message.getMessageId());
+      serviceBusReceiverClient.abandon(message);
```
rtitle, line 114: Does `abandon` cause the message to get reprocessed by the queue? (which is the semantics we want, I think)

stairway-azure/src/main/java/bio/terra/stairway/azure/AzureServiceBusQueue.java
```diff
@@ -0,0 +1,240 @@
...
+    // be re-thrown
+    List<ServiceBusReceivedMessage> messages =
+        serviceBusReceiverClient.receiveMessages(maxMessages).stream().toList();
+
+    for (ServiceBusReceivedMessage message : messages) {
+      ProcessMessage(processFunction, message);
+    }
+  }
+
+  private void ProcessMessage(
```
rtitle: Nit: camelCase `processMessage` :)

stairway-azure/src/main/java/bio/terra/stairway/azure/AzureServiceBusQueue.java
```diff
@@ -0,0 +1,240 @@
...
+        serviceBusClientBuilder
+            .receiver()
+            .topicName(builder.topicName)
+            .subscriptionName(builder.subscriptionName)
+            .receiveMode(ServiceBusReceiveMode.PEEK_LOCK)
+            .disableAutoComplete()
+            .maxAutoLockRenewDuration(builder.maxAutoLockRenewDuration)
+            .buildClient();
+
+    serviceBusSenderClient =
```
rtitle, line 74: Do we want a `shutdown` method to close these clients & their connections?

rtitle (commented):
A few comments.. overall approach looks good. Though I'm new to both Stairway code and Service Bus.

dvoet (approved):

stairway-azure/src/main/java/bio/terra/stairway/azure/AzureServiceBusQueue.java
giventocode: Correct, it returns the message to the queue.

stairway-azure/src/main/java/bio/terra/stairway/azure/AzureServiceBusQueue.java
giventocode: Doing too much C# lately... :).

stairway-azure/src/main/java/bio/terra/stairway/azure/AzureServiceBusQueue.java
giventocode: Yeah, I wondered about this as well. The GCP implementation has a shutdown method, however it is not part of the interface. Now that you mention it and just to be safe, I'll add the method for consistency's sake.

rtitle (approved):

PR #115, opened 2024-02-26
[PF-2987] Add dependabot for 3rd party library upgrades, remove lockfiles and spotbugs

cahrens:
Same configuration as in terra-common-library.

Also removes lockfiles and spotbugs, similar to what we have done in our other libraries (for example, https://github.com/DataBiosphere/terra-common-lib/pull/148 and https://github.com/DataBiosphere/terra-common-lib/pull/147).


stairctl/gradle.lockfile
```diff
@@ -1,140 +0,0 @@
-# This is a Gradle generated file for dependency locking.
```
jsotobroad, line 1: not that I"m complaining but why the removal of hte .lockfiles?

.github/dependabot.yml
```diff
@@ -1,11 +1,24 @@
...
+          - "minor"
+          - "patch"
     schedule:
-      interval: "weekly"
+      interval: "monthly"
+      time: "06:00"
+      timezone: "America/New_York"
+    target-branch: "develop"
+    reviewers:
+      - "@DataBiosphere/platform-foundation-codeowners"
```
jsotobroad, line 19: just double checking that these reviewers are the group we want for this (out of ignorance on my part 😬)

cahrens: Yes, we talked about this today at the "round table" meeting. It might not be the perfect group for reviews, but we don't wish to maintain a second group.

stairctl/gradle.lockfile
cahrens: We've been removing lockfiles in all of our repos to make dependency upgrades easier.

cahrens:
I don't understand why this isn't kicking off any tests (given the source code changes).

cahrens:
It looks like tests are not triggered on commits. I am putting this PR into draft until we get unit tests running.

stairctl/src/main/java/bio/terra/stairctl/commands/Output.java
```diff
@@ -10,13 +10,13 @@
...
 import java.util.List;
 import java.util.Optional;
-import javax.annotation.Nullable;
 import org.apache.commons.lang3.ClassUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.lang.Nullable;
```
cahrens, line 19: `javax.annotation.Nullable` was coming from spotbugs (https://www.javadoc.io/doc/com.google.code.findbugs/jsr305/latest/javax/annotation/Nullable.html).

There seem to be lots of options for @Nullable, so I went with a package that was already available with our current dependencies. We do use this package in some of our other repos. https://github.com/search?q=org%3ADataBiosphere%20org.springframework.lang.Nullable&type=code

jsotobroad (commented):
It sounds like we're not sure if we want to remove spotbugs yet since sonarqube isnt in this repo.  Would it be worthwhile to separate the dependabot changes from the spotbugs stuff to get that in?

The changes look fine overall.

cahrens:
> It sounds like we're not sure if we want to remove spotbugs yet since sonarqube isnt in this repo. Would it be worthwhile to separate the dependabot changes from the spotbugs stuff to get that in?
> 
> The changes look fine overall.

This, @jsotobroad . I agree that I could pull out the spotbugs removal, but first I'm going to reach out to infosec folks and see if we can just get sonarqube running on this repo.

cahrens:
> > It sounds like we're not sure if we want to remove spotbugs yet since sonarqube isnt in this repo. Would it be worthwhile to separate the dependabot changes from the spotbugs stuff to get that in?
> > The changes look fine overall.
> 
> This, @jsotobroad . I agree that I could pull out the spotbugs removal, but first I'm going to reach out to infosec folks and see if we can just get sonarqube running on this repo.

@jsotobroad , @TomConner is OK with us going ahead and removing spotbugs. So this PR is now ready for review.

https://broadinstitute.slack.com/archives/CADU7L0SZ/p1709132894654009

.github/workflows/build-and-test.yaml
```diff
@@ -50,7 +48,7 @@ jobs:
           cache: 'gradle'
       - name: Run tests
         run: ./gradlew test
-      
+
       # TODO: Work with AppSec to get Sonar scans setup for this repo
```
TomConner, line 52: 👋

okotsopoulos (approved):
Thanks for taking this on! I have some Stairway work in flight and will be really happy to see the lockfiles go away :)

dvoet (approved):

stairctl/src/main/java/bio/terra/stairctl/commands/Output.java
pshapiro4broad: I think a better fix would be to use `jakarta.annotation.Nullable`, as it doesn't require a framework dependency.

cahrens: It didn't seem that was available with our current dependencies/framework, but perhaps IntelliJ was lying to me?

pshapiro4broad: Yes, it might need an additional dependency like https://mvnrepository.com/artifact/jakarta.annotation/jakarta.annotation-api

cahrens: Right, I started looking into adding that, but then punted since we use `org.springframework.lang.Nullable` in some of our other repos.

PR #116, opened 2024-02-27
PF-2033, PF-2989: Trigger Unit Tests On PR Commits

mikeflinn:
Adds a github actions workflow to add CI that will run all the unit tests and perform a build of the library upon PR commits and merges to develop for this repo.

Some future improvements we may want to consider adding are asking AppSec to setup Sonar scans for this repo and adding Workflow failure to slack reporting.

### DIFF
--- a/.github/workflows/unitTest.yaml
+++ b/.github/workflows/unitTest.yaml
@@ -1,0 +1,34 @@
+name: Run Unit Tests
+
+on:
+  push:
+    branches: [ develop ]
+    paths-ignore: [ '*.md' ]
+  pull_request:
+    branches: [ '**' ]
+
+jobs:
+  run-unit-tests:
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:13
+        env:
+            # Default values stairway expects based on DEVELOPMENT.md
+            POSTGRES_PASSWORD: stairwaypw
+            POSTGRES_USER: stairwayuser
+            POSTGRES_DB: stairwaylib
+        ports: [ "5432:5432" ]
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Set up JDK
+        uses: actions/setup-java@v3
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+          cache: 'gradle'
+      - name: Run tests
+        run: ./gradlew --build-cache test
+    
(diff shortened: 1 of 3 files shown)


.github/workflows/unitTest.yaml
```diff
@@ -0,0 +1,34 @@
...
+        with:
+          fetch-depth: 0
+      - name: Set up JDK
+        uses: actions/setup-java@v3
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+          cache: 'gradle'
+      - name: Run tests
+        run: ./gradlew --build-cache test
```
pshapiro4broad: I know we do this, but I don't think `--build-cache` is required. https://github.com/actions/setup-java?tab=readme-ov-file#caching-gradle-dependencies

.github/workflows/unitTest.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+on:
+  push:
+    branches: [ develop ]
+    paths-ignore: [ '*.md' ]
+  pull_request:
+    branches: [ '**' ]
+
+jobs:
+  run-unit-tests:
```
pshapiro4broad: Maybe this should use the CI workflow from the template project? https://github.com/DataBiosphere/terra-java-project-template/blob/main/.github/workflows/build-and-test.yml

The template workflow runs build separately from test.

Maybe that can wait though; we'd also want to get sonar working with coverage.

mikeflinn: This was just a quick experiment to see if I could get the tests to run and pass at all. Agree it should use the standard CI workflow

mikeflinn: After looking some more I don't think it makes sense to use an exact copy paste of the template repo ci workflow here. It includes some additional logic which isn't relevant for stairway or other libraries such as running integration tests with test runner and building a docker image and running scans against it etc..

Running a build of the binary and the unit test suite in parallel does make sense to me though.

The failure notifications to slack would also make sense to me to add here, but I think that can be done separately since it requires a terraform change to setup the secrets.

mikeflinn: Also agree we should have sonar here but I think that needs to be setup separately by infosec.

okotsopoulos (approved):
Thanks so much for doing this!

Do we also want to run connected tests on PRs?  (Even if so, I'd say that we want to handle that separately.)

pshapiro4broad (approved):
🎉

PR #117, opened 2024-02-28
[WOR-1544] Populate MDC for flight and its steps

okotsopoulos:
https://broadworkbench.atlassian.net/browse/WOR-1544

Wrap flight thread execution so that flight threads are context-aware.

- Logs emitted during a flight thread now have the context from the calling thread, along with flight-specific context (flight ID and class).
- Logs emitted during the execution of a step additionally include step-specific context (step class, direction, and index). 

Handles the case where a flight thread spawns a child flight by intentionally cleaning any leftover step-specific context from the calling thread.


stairway/src/main/java/bio/terra/stairway/impl/FlightRunner.java
```diff
@@ -144,11 +144,9 @@ private FlightStatus fly() throws InterruptedException {
       // the UNDO in the first place.
       flightDao.step(flightContext);
       logger.error(
-          "DISMAL FAILURE: non-retry-able error during undo. Flight: {}({}) Step: {}({})",
-          flightContext.getFlightId(),
-          flightContext.getFlightClassName(),
-          flightContext.getStepIndex(),
-          flightContext.getStepClassName());
+          "{} (index {}) experienced DISMAL FAILURE: non-retryable error on undo",
```
okotsopoulos, line 147: Drive-by: now that we include flight ID and class on all flight logs in the MDC, we can trim down the dismal failure log line.

I also reformatted to lead with the failing step class based off of Stairway retro feedback that all of the useful information on a log line is at the end.  Do we like this better?  Unfortunately we don't get step context in MDC for free at this point -- we're outside of step execution.

okotsopoulos:
Got into a weird Git state, recreating.

PR #125, opened 2024-02-29
[WOR-1544] #major Populate MDC for flight and its steps

okotsopoulos:
https://broadworkbench.atlassian.net/browse/WOR-1544

Centralizing work done in https://github.com/DataBiosphere/terra-workspace-manager/pull/1626, wrap flight thread execution so that flight threads are context-aware.

- Logs emitted during a flight thread now have the **context from the calling thread**, along with **flight-specific context** (flight ID and class).
- Logs emitted during the execution of a step additionally include **step-specific context** (step class, direction, and index). 

Handles the case where a flight thread spawns a child flight by intentionally cleaning any leftover step-specific context from the calling thread.

**In Action**

I published my Stairway branch locally, pulled it in as a WSM dependency, removed the existing WSM code which modifies the MDC, deployed it to my Bee, and hit an endpoint that triggers a stairway flight.

The resulting logs have their MDC populated as described in this PR: https://cloudlogging.app.goo.gl/T9vdJ86RXRoa117VA
- Calling thread's context over the entirety of the flight (that's where `jsonPayload.requestId` comes from)
- Flight-specific context over the entirety of the flight
- Step-specific context over the entirety of each step

A `jsonPayload` for a log emitted during a step:
Screenshot 2024-02-29 at 7 44 47 PM

All of the logs for the request, with `flightId`, `flightStepNumber`, and `flightStepDirection` added to the summary line:
Screenshot 2024-02-29 at 7 46 57 PM

(As an aside, it's clear to me that these log lines can be substantially trimmed.  Do we need the full formatted summary anymore if we have this info in the MDC, and can pull it in the summary line?)

**How can my service get this new functionality?**

Once this PR is merged, I'll update the dependency in TCL, which is where most services get Stairway from.  At any rate, downstream services will need to update a dependency.

If downstream services presently modify their MDCs themselves, they will be able to remove that code in their service and let Stairway take care of it.


stairway/src/main/java/bio/terra/stairway/impl/FlightRunner.java
```diff
@@ -144,11 +144,9 @@ private FlightStatus fly() throws InterruptedException {
       // the UNDO in the first place.
       flightDao.step(flightContext);
       logger.error(
-          "DISMAL FAILURE: non-retry-able error during undo. Flight: {}({}) Step: {}({})",
-          flightContext.getFlightId(),
-          flightContext.getFlightClassName(),
-          flightContext.getStepIndex(),
-          flightContext.getStepClassName());
+          "{} (index {}) experienced DISMAL FAILURE: non-retryable error on undo",
```
okotsopoulos, line 147: Drive-by: now that we include flight ID and class on all flight logs in the MDC, we can trim down the dismal failure log line.

I also reformatted to lead with the failing step class based off of Stairway retro feedback that all of the useful information on a log line is at the end. Do we like this better? Unfortunately we don't get step context in MDC for free at this point -- we're outside of step execution.

README.md
```diff
@@ -93,6 +93,16 @@ Stairway is designed to provide atomic operations for one instance of one servic
 global or cross-service state. Therefore, it is up to the application or service to implement concurrency control on 
 its objects.
 
+### Context Awareness and Logs
+Stairway leverages the underlying logging system's mapped diagnostic context (MDC) if available.
+
+MDC manages contextual information on a per-thread basis: as Stairway submits a Flight for processing,
+it **passes along the MDC of the calling thread** so that context isn't lost.  It also further populates
```
okotsopoulos, line 100: Propagating the MDC from the calling thread makes me want to designate this as a minor version bump at minimum.  Default behavior in this repo is patch bumps.

Reviewers, WDYT?

buildSrc/src/main/groovy/stairway.java-conventions.gradle
```diff
@@ -16,8 +16,7 @@ dependencies {
     implementation group: 'org.slf4j', name: 'slf4j-api', version: '1.7.36'
 
     // For testing
-    testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api', version: '5.9.0'
-    testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-engine', version: '5.9.0'
+    testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter', version: '5.10.2'
```
okotsopoulos, line 19: Needed this expanded dependency for parameterized tests, and upgraded while in the neighborhood.

README.md
cahrens: I agree with a more significant version bump. Though with all the dependency upgrades underway, I was already thinking we would do a major version bump very soon. Perhaps just do a major bump with this change?

stairway/src/main/java/bio/terra/stairway/impl/MdcHelper.java
```diff
@@ -0,0 +1,76 @@
...
+  /**
+   * @return the flightRunner modified to propagate the MDC from the calling thread and
+   *     flight-specific context to the child thread spawned to run the flight.
+   */
+  static Runnable withMdcAndFlightContext(Runnable flightRunner, FlightContext flightContext) {
+    // Save the calling thread's context
+    Map<String, String> contextMap = MDC.getCopyOfContextMap();
+    return () -> {
+      // Any leftover context on the thread will be fully overwritten:
+      MDC.setContextMap(Optional.ofNullable(contextMap).orElse(Map.of()));
```
cahrens: I'm having trouble understanding this line. We are getting a copy of the context map at line 35, so why isn't line 38 effectively a no-op?

stairway/src/main/java/bio/terra/stairway/impl/MdcHelper.java
```diff
@@ -0,0 +1,76 @@
...
+      // If the calling thread's context contains flight and step context from a parent flight, this
+      // will be overwritten below:
+      addFlightContextToMdc(flightContext);
+      removeStepContextFromMdc(flightContext);
```
cahrens: I don't understand removing step context here.

stairway/src/test/java/bio/terra/stairway/impl/MdcHelperTest.java
```diff
@@ -0,0 +1,139 @@
...
+        new TestFlightContext()
+            .flightId(FLIGHT_ID)
+            .flightClassName(FLIGHT_CLASS)
+            .stepIndex(STEP_INDEX)
+            .direction(STEP_DIRECTION)
+            .stepClassName(STEP_CLASS);
+  }
+
+  static Stream<Map<String, String>> initialContext() {
+    return Stream.of(null, Map.of(), FOO_BAR);
```
cahrens, line 52: Isn't this just returning a single map? If so, I don't see the point of making the tests parameterized.

README.md
okotsopoulos: Good idea. I'll make it a major one.

stairway/src/main/java/bio/terra/stairway/impl/MdcHelper.java
okotsopoulos: Ah, great question.  Stairway spawns new threads for flight execution, so this Runnable within may be happening on a new thread without the prior context.

But this tells me that this code would be better organized if it's added to the Stairway thread pool implementation instead. I'll make that change.

dvoet (approved):

stairway/src/main/java/bio/terra/stairway/impl/MdcHelper.java
okotsopoulos: When I added flight and step context to the MDC in WSM's MdcHook, I didn't account for subflights that were kicked off within a parent flight's step.  The initial logs emitted by subflights before the first step errantly included step-specific context from the calling flight, interspersed with flight-specific context from the new subflight.  https://github.com/DataBiosphere/terra-workspace-manager/pull/1636

What changes would help make this clear?

cahrens: Ah, I get it now.

README.md
okotsopoulos: Added #major to the PR title, which will get pulled into the commit message.

stairway/src/main/java/bio/terra/stairway/impl/MdcHelper.java
okotsopoulos: Hi @cahrens , I reorganized this code to live in `StairwayThreadPool` and feel like it makes more sense now, thanks for the original comment.  Would you mind taking another look?

stairway/src/main/java/bio/terra/stairway/impl/StairwayThreadPool.java
```diff
@@ -29,10 +33,45 @@ int getQueuedFlights() {
...
+    // will be overwritten below:
+    MdcUtils.addFlightContextToMdc(flightContext);
+    MdcUtils.removeStepContextFromMdc(flightContext);
+  }
+
   protected void beforeExecute(Thread t, Runnable r) {
-    super.beforeExecute(t, r);
     int active = activeTasks.incrementAndGet();
     logger.debug("before: " + active);
+    super.beforeExecute(t, r);
```
okotsopoulos, line 74: Reading the docstring on this method, it's recommended to call the super at the end to allow for potential nesting.

(I also read the docstring on `afterExecute`, and the super call at the beginning matches the recommendation.)

.github/workflows/build-and-test.yaml
```diff
@@ -47,7 +47,7 @@ jobs:
           distribution: 'temurin'
           cache: 'gradle'
       - name: Run tests
-        run: ./gradlew test
+        run: ./gradlew test --scan
```
cahrens, line 50: What does this do?

stairway/src/main/java/bio/terra/stairway/impl/FlightRunner.java
```diff
@@ -242,6 +240,7 @@ private StepResult stepWithRetry() throws InterruptedException, StairwayExecutio
     // Retry loop
     do {
       try {
+        MdcUtils.addStepContextToMdc(flightContext);
```
cahrens, line 243: Ah, it makes a lot more sense here-- thanks!

stairway/src/main/java/bio/terra/stairway/impl/MdcUtils.java
```diff
@@ -0,0 +1,74 @@
...
+  /**
+   * Null-safe utility method for overwriting the current thread's MDC.
+   *
+   * @param context to set as MDC, if null then MDC will be cleared.
+   */
+  static void overwriteContext(Map<String, String> context) {
+    if (context == null) {
+      MDC.clear();
+    } else {
+      MDC.setContextMap(context);
```
cahrens, line 36: If there was already a context map, does this completely clear it (as opposed to merging)? 

I was just wondering if the following would be more robust/clearer:

```
MDC.clear();
if (context != null) {
  MDC.setContextMap(context);
}

**Review by cahrens — APPROVED** [review]

File: .github/workflows/build-and-test.yaml (hunk 8 above)
Comment (okotsopoulos) [reply .github/workflows/build-and-test.yaml:50]: This generates a Gradle build scan as part of test output, which makes debugging tests a lot easier.  Example from the most recent unit test run:

```
Publishing build scan...
https://gradle.com/s/gu4jckjm4mqzq
```

stairway/src/main/java/bio/terra/stairway/impl/MdcUtils.java
okotsopoulos: They are functionally the same per MDC's documentation, but I agree that your version is clearer.  Will update.

PR #129, opened 2024-03-01
[WOR-1531] Upgrade postgresql to 42.7.2

okotsopoulos:
https://broadworkbench.atlassian.net/browse/WOR-1531

Addresses https://nvd.nist.gov/vuln/detail/CVE-2024-1597.  (This minor version bump is currently stuck in a [large Dependabot grouped update](https://github.com/DataBiosphere/stairway/pull/127), so breaking it out.)

### DIFF
--- a/stairctl/build.gradle
+++ b/stairctl/build.gradle
@@ -28,5 +28,5 @@ dependencies {
     implementation group: 'org.springframework.shell', name: 'spring-shell-starter', version: '2.1.1'
 
     // Database
-    runtimeOnly group: 'org.postgresql', name: 'postgresql'
+    runtimeOnly group: 'org.postgresql', name: 'postgresql', version: '42.7.2'
 }
--- a/stairctl/build.gradle
+++ b/stairctl/build.gradle
@@ -26,7 +26,4 @@ dependencies {
     // Spring
     implementation group: 'org.springframework.boot', name: 'spring-boot-starter' // version from spring-boot-plugin
     implementation group: 'org.springframework.shell', name: 'spring-shell-starter', version: '2.1.1'
-
-    // Database
-    runtimeOnly group: 'org.postgresql', name: 'postgresql', version: '42.7.2'
 }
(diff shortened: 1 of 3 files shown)


stairctl/build.gradle
```diff
@@ -28,5 +28,5 @@ dependencies {
     implementation group: 'org.springframework.shell', name: 'spring-shell-starter', version: '2.1.1'
 
     // Database
-    runtimeOnly group: 'org.postgresql', name: 'postgresql'
+    runtimeOnly group: 'org.postgresql', name: 'postgresql', version: '42.7.2'
```
okotsopoulos: Explicitly setting the version here was necessary for Gradle's dependencyInsight task to show that we've moved past the vulnerability.

tlangs (approved):

stairctl/build.gradle
pshapiro4broad: Could updating the spring dependency plugin fix this? I think that's where this dependency is coming from.

okotsopoulos: That didn't fix this but on further reflection I think we'd prefer that stairctl use the same Postgres as the main stairway project, rather than following the Spring Boot dependency plug-in.

I'm inclined to move this to `stairway.java-conventions` to enforce the version throughout, rather than defining here and in `stairway/build.gradle`.  All other projects currently depend on `stairway` project anyway, so they're pulling in the dependency already.  But I'm not solid on this.

@pshapiro4broad do you have a better idea?  Happy to catch up async.

pshapiro4broad (approved):
Discussed directly in slack

stairctl/build.gradle
pshapiro4broad: For the (GitHub) record, I think this dependency can be moved into the java conventions subproject to avoid the duplicate version pinning. Also, I think the dependency block in the root build.gradle is a vestige of adding version lock support and can be removed if we're not doing version locking anymore.

okotsopoulos: Thanks for chatting about this -- I've moved this dependency to stairway.java-conventions.

PR #137, opened 2024-03-26
[WOR-1591] Update "isActive" to include all states

cahrens:
It appears that `isActive` was implemented before the introduction of a number of new states. Therefore its check for only RUNNING as being "active" meant that other non-terminal states did not get considered "active". I am fixing this, but also encouraging that we do _not_ use `waitForFlight` in our production code. To address [WOR-1591](https://broadworkbench.atlassian.net/browse/WOR-1591) I will be changing the WSM usages of `waitForFlight` to the more robust WSM `FlightUtils.waitForFlightCompletion` implementation.

[WOR-1591]: https://broadworkbench.atlassian.net/browse/WOR-1591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/stairway/src/main/java/bio/terra/stairway/FlightState.java
+++ b/stairway/src/main/java/bio/terra/stairway/FlightState.java
@@ -107,7 +107,7 @@ public class FlightState {
   /** Returns true if the flight is not in a terminal state. */
   public boolean isActive() {
     boolean complete =
-        (flightStatus == FlightStatus.RUNNING
+        (flightStatus == FlightStatus.ERROR
             || flightStatus == FlightStatus.FATAL
             || flightStatus == FlightStatus.SUCCESS);
     return !complete;
--- a/stairway/src/main/java/bio/terra/stairway/FlightState.java
+++ b/stairway/src/main/java/bio/terra/stairway/FlightState.java
@@ -104,7 +104,12 @@ public class FlightState {
     return progressMeters;
   }
 
+  /** Returns true if the flight is not in a terminal state. */
   public boolean isActive() {
-    return (flightStatus == FlightStatus.RUNNING);
+    boolean complete =
+        (flightStatus == FlightStatus.RUNNING
+            || flightStatus == FlightStatus.FATAL
+            || flightStatus == FlightStatus.SUCCESS);
+    return !complete;
   }
 }
--- a/stairway/src/main/java/bio/terra/stairway/impl/StairwayImpl.java
+++ b/stairway/src/main/java/bio/terra/stairway/impl/StairwayImpl.java
@@ -497,6 +497,7 @@ public class StairwayImpl implements Stairway {
    * @throws FlightNotFoundException flight id does not exist
    * @throws FlightWaitTimedOutException if interrupted or polling interval expired
    * @throws InterruptedException on shutdown while waiting for flight completion
+   * @deprecated
    */
   public FlightState waitForFlight(String flightId, Integer pollSeconds, Integer pollCycles)
       throws StairwayException,
--- a/stairway/src/main/java/bio/terra/stairway/impl/StairwayImpl.java
+++ b/stairway/src/main/java/bio/terra/stairway/impl/StairwayImpl.java
@@ -133,8 +133,11 @@ public class StairwayImpl implements Stairway {
    */
   public List<String> initialize(
       DataSource dataSource, boolean forceCleanStart, boolean migrateUpgrade)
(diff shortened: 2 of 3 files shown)


stairway/src/main/java/bio/terra/stairway/FlightState.java
```diff
@@ -104,7 +104,12 @@ public ProgressMeterReader getProgressMeters() {
     return progressMeters;
   }
 
+  /** Returns true if the flight is not in a terminal state. */
   public boolean isActive() {
```
cahrens, line 108: Using the same implementation as https://github.com/DataBiosphere/terra-workspace-manager/blob/0131b1e681bc8125428ec1aebf9030ee7e375b95/service/src/main/java/bio/terra/workspace/common/utils/FlightUtils.java#L274

stairway/src/main/java/bio/terra/stairway/impl/StairwayImpl.java
```diff
@@ -133,8 +133,11 @@ public StairwayImpl(StairwayBuilder builder) throws StairwayExecutionException {
    */
   public List<String> initialize(
       DataSource dataSource, boolean forceCleanStart, boolean migrateUpgrade)
-      throws StairwayShutdownException, DatabaseOperationException, MigrateException,
-          StairwayException, InterruptedException {
+      throws StairwayShutdownException,
```
cahrens, line 136: spotless and documentation changes only in this file

stairway/src/main/java/bio/terra/stairway/FlightState.java
okotsopoulos: I see this is a public method.  When you make changes in WSM, if we still need to use this implementation could we instead call the Stairway method?

stairway/src/main/java/bio/terra/stairway/impl/StairwayImpl.java
```diff
@@ -484,10 +497,14 @@ public void deleteFlight(String flightId, boolean forceDelete)
    * @throws FlightNotFoundException flight id does not exist
    * @throws FlightWaitTimedOutException if interrupted or polling interval expired
    * @throws InterruptedException on shutdown while waiting for flight completion
+   * @deprecated
    */
   public FlightState waitForFlight(String flightId, Integer pollSeconds, Integer pollCycles)
```
okotsopoulos, line 502: This is OOS, but I don't love how this takes in `pollSeconds` rather than a `Duration` to poll for.  Tests that use this method end up taking a lot longer to run on account of the default 10s sleeps.  But, hopefully this is a moot point given the deprecation.

stairway/src/main/java/bio/terra/stairway/FlightState.java
cahrens: I'm not following. I don't believe we will call this method at all.

okotsopoulos: Also, would it be feasible to write a unit test for this, given that this is a bugfix?

okotsopoulos (approved):

stairway/src/main/java/bio/terra/stairway/FlightState.java
cahrens: Ohhh, you are suggesting we call this method from WSM? Sure, I guess we could if we have updated WSM to use the new version of Stairway. I'm not sure how strongly I feel about it, but I could be convinced.

cahrens: Sure, I will write a unit test for this. I was thinking of this more as an example implementation, but now am coming around to it being possibly more widely used.

okotsopoulos: > Ohhh, you are suggesting we call this method from WSM? Sure, I guess we could if we have updated WSM to use the new version of Stairway. I'm not sure how strongly I feel about it, but I could be convinced.

Yes, that's what I was getting at, rather than duplicating logic between services (it does seem like it makes sense to live in Stairway).

cahrens: Added test-- https://github.com/DataBiosphere/stairway/pull/137/commits/9312dd2211f8ae08a111ef40493964f35cbb233b

sergiygetlin (approved):

PR #148, opened 2024-06-03
PF-2987: minor patch dependencies

davidangb:
supersedes Dependabot's #147, which has build failures.

The Spring Boot and Spring Shell versions should stay in sync, but that other PR uses divergent versions and that is the cause of the build failures.

This PR is branched from that other PR but has an extra commit 5d887bed7512831b07b4641eec3bccb7eb36a50d that fixes the Boot/Shell versions.

NOTE: Dependabot could very well try to upgrade Spring Boot again in its next pass; I am not trying to prevent that in this PR.

### DIFF
--- a/stairctl/build.gradle
+++ b/stairctl/build.gradle
@@ -2,8 +2,8 @@ plugins {
     id 'application'
     id 'stairway.java-conventions'
 
-    id 'io.spring.dependency-management' version '1.1.4'
-    id 'org.springframework.boot' version '3.2.5'
+    id 'io.spring.dependency-management' version '1.1.5'
+    id 'org.springframework.boot' version '3.3.0'
 }
 
 version = gradle.version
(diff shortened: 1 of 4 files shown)


cahrens (approved):

stairctl/build.gradle
```diff
@@ -2,7 +2,8 @@ plugins {
     id 'application'
     id 'stairway.java-conventions'
 
-    id 'io.spring.dependency-management' version '1.1.4'
+    id 'io.spring.dependency-management' version '1.1.5'
+    // Spring Boot and Spring Shell versions should stay in sync
```
cjllanwarne, line 6: Should or Must?

If Must, can we make this a variable instead of copy/pasting the same value?

cjllanwarne (approved):

stairctl/build.gradle
davidangb: I honestly don't know if it's should or must. These two versions - 3.2.5 and 3.3.0 - cause a problem. I don't know if other version combinations also cause a problem, so I'm kicking the can down the road and being noncommittal

PR #149, opened 2024-06-28
[DCJ-457] Stairway can be built with a provided ThreadPoolTaskExecutor

okotsopoulos:
## Addresses
Partially addresses https://broadworkbench.atlassian.net/browse/DCJ-457

PRs will need to be merged in the following order:
1. This PR
2. https://github.com/DataBiosphere/terra-common-lib/pull/187
3. A forthcoming TDR PR to update its TCL version

## Summary of changes

This PR allows calling services to bring their own instrumented executor to serve as their Stairway's threadpool.

If unspecified in `StairwayBuilder`, a `DefaultThreadPoolTaskExecutor` will be constructed and used.

`DefaultThreadPoolTaskExecutor` is a public class: calling services may elect to instantiate it themselves (e.g. a Spring Boot service would register this as a Bean for automatic actuator instrumentation).

## Testing Strategy

Expanded unit tests.

Verified end-to-end behavior manually:
- Publishing Stairway and TCL to my local Maven repository
- Booting up TDR, with and without `maxParallelFlights` specified
- Observed that `stairwayExecutor` metrics were automatically flowing through TDR's Actuator endpoint
- Initiated a Stairway flight which succeeded.  Executor occupancy was reflected in actuator metrics.

Example where TDR configured its `maxParallelFlights` to be 120, has one actively running flight, and one completed flight:
```
(base) okotsopo@wm111-e35 jade-data-repo % curl http://localhost:8080/actuator/prometheus | grep 'stairwayExecutor'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 87022  100 8executor_pool_core_threads{name="stairwayExecutor",} 120.0   0
7022    0     0  10.1M      0 --:--:-- --:--:-- --:--:-- 10.3M
executor_queue_remaining_tasks{name="stairwayExecutor",} 2.147483647E9
executor_completed_tasks_total{name="stairwayExecutor",} 1.0
executor_pool_size_threads{name="stairwayExecutor",} 2.0
executor_queued_tasks{name="stairwayExecutor",} 0.0
executor_pool_max_threads{name="stairwayExecutor",} 120.0
executor_active_threads{name="stairwayExecutor",} 1.0
```

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -54,12 +54,15 @@ Gradle makes this easy with a `mavenLocal` target for publishing and loading pac
 
    ```
    # terra-workspace-manager/build.gradle
-   
+
+   // If true, search local repository (~/.m2/repository/) first for dependencies.
+   def useMavenLocal = true
    repositories {
-     mavenLocal()
-     mavenCentral()
-     ...
-   }
+      if (useMavenLocal) {
+          mavenLocal() // must be listed first to take effect
+      }
+      mavenCentral()
+      ...
    ```
 
 That's it! Your service should pick up locally-published changes. If your changes involved bumping
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -35,6 +35,36 @@ match the values in the sql file above:
 For folks working on Terra, the Stairway configuration is embedded within the component
 configuration, so these steps are included in component developer setup.
 
+## Local Publishing
+When working on this library, it is often helpful to be able to quickly test out changes
+in the context of a service repo (e.g. `terra-workspace-manager` or `terra-resource-buffer`)
+running a local server.
+
+Gradle makes this easy with a `mavenLocal` target for publishing and loading packages:
+
+1. Publish from Stairway to your machine's local Maven cache.
+
+   ```
+   ./gradlew publishToMavenLocal
+   ```
+
+   Your package will be in `~/.m2/repository`.
+2. From the service repo, add `mavenLocal()` to the _first_ repository location
+   build.gradle file (e.g. before `mavenCentral()`).
+
+   ```
+   # terra-workspace-manager/build.gradle
+   
+   repositories {
+     mavenLocal()
+     mavenCentral()
+     ...
+   }
+   ```
+
+That's it! Your service should pick up locally-published changes. If your changes involved bumping
+this library's version, be careful to update version numbers accordingly.
+
 ## SourceClear
 
 [SourceClear](https://srcclr.github.io) is a static analysis tool that scans a project's Java
--- a/stairway/src/main/java/bio/terra/stairway/DefaultThreadPoolTaskExecutor.java
+++ b/stairway/src/main/java/bio/terra/stairway/DefaultThreadPoolTaskExecutor.java
@@ -1,0 +1,31 @@
+package bio.terra.stairway;
+
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
+
+public class DefaultThreadPoolTaskExecutor extends ThreadPoolTaskExecutor {
+
+  static final int DEFAULT_MAX_PARALLEL_FLIGHTS = 20;
+
+  /**
+   * An initialized {@link ThreadPoolTaskExecutor}.
+   *
+   * @param maxParallelFlights the desired pool size, honored if a positive integer, defaults to
+   *     DEFAULT_MAX_PARALLEL_FLIGHTS otherwise.
+   */
+  public DefaultThreadPoolTaskExecutor(Integer maxParallelFlights) {
+    super();
+    int poolSize = getPoolSize(maxParallelFlights);
+    super.setCorePoolSize(poolSize);
+    super.setMaxPoolSize(poolSize);
+    super.setKeepAliveSeconds(0);
+    super.setThreadNamePrefix("stairway-thread-");
+    super.initialize();
+  }
+
+  private int getPoolSize(Integer maxParallelFlights) {
+    if (maxParallelFlights == null || maxParallelFlights <= 0) {
+      return DEFAULT_MAX_PARALLEL_FLIGHTS;
+    }
+    return maxParallelFlights;
+  }
+}
--- a/stairway/src/main/java/bio/terra/stairway/StairwayBuilder.java
+++ b/stairway/src/main/java/bio/terra/stairway/StairwayBuilder.java
@@ -31,8 +31,8 @@ public class StairwayBuilder {
   private Duration completedFlightRetention;
 
   /**
-   * Determines the size of the thread pool used for running Stairway flights. Default is
-   * established in {@link DefaultThreadPoolTaskExecutor} (20 at this moment).
+   * Determines the size of the thread pool used for running Stairway flights. Default is {@value
+   * DefaultThreadPoolTaskExecutor#DEFAULT_MAX_PARALLEL_FLIGHTS}.
    *
    * @param maxParallelFlights maximum parallel flights to run
    * @return this
--- a/stairway/src/main/java/bio/terra/stairway/StairwayBuilder.java
+++ b/stairway/src/main/java/bio/terra/stairway/StairwayBuilder.java
@@ -5,6 +5,7 @@ import bio.terra.stairway.impl.StairwayImpl;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
 
 /**
  * This builder class is the way to constructing a Stairway instance. For example,
@@ -21,6 +22,7 @@ public class StairwayBuilder {
   private final List<StairwayHook> stairwayHooks = new ArrayList<>();
   private Integer maxParallelFlights;
   private Integer maxQueuedFlights;
+  private ThreadPoolTaskExecutor executor;
   private Object applicationContext;
   private ExceptionSerializer exceptionSerializer;
   private String stairwayName;
@@ -30,7 +32,7 @@ public class StairwayBuilder {
 
   /**
    * Determines the size of the thread pool used for running Stairway flights. Default is
-   * DEFAULT_MAX_PARALLEL_FLIGHTS (20 at this moment)
+   * established in {@link DefaultThreadPoolTaskExecutor} (20 at this moment).
    *
    * @param maxParallelFlights maximum parallel flights to run
    * @return this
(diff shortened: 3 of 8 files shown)


okotsopoulos:
(Only marked as "ready for review" to see if that helps with checks not running… but it seems that GitHub is experiencing issues: https://www.githubstatus.com/incidents/9vwllhs2w1kj)

stairway/src/main/java/bio/terra/stairway/impl/StairwayThreadPool.java
```diff
@@ -3,34 +3,15 @@
...
-  }
-
-  int getActiveFlights() {
-    return activeTasks.get();
-  }
+  private final ThreadPoolTaskExecutor executor;
 
-  int getQueuedFlights() {
-    return getQueue().size();
+  StairwayThreadPool(ThreadPoolTaskExecutor executor) {
```
pshapiro4broad, line 13: I like the idea of moving the thread pool to a field rather than a base class. Is there a reason why you didn't continue to use `ThreadPoolExecutor` here? That way a user doesn't need to use a spring type to provide a custom thread pool. I believe the spring types are compatible with the java ones, so using `ThreadPoolExecutor` would still allow clients to use `ThreadPoolTaskExecutor` if they wanted to.

pshapiro4broad: Reading on, it looks like this uses spring to track the number of active threads rather than doing it ourselves. I guess there's no harm in adding more spring dependencies here as we already use spring for all the java apps.

DEVELOPMENT.md
```diff
@@ -35,6 +35,36 @@ match the values in the sql file above:
...
+
+   Your package will be in `~/.m2/repository`.
+2. From the service repo, add `mavenLocal()` to the _first_ repository location
+   build.gradle file (e.g. before `mavenCentral()`).
+
+   ```
+   # terra-workspace-manager/build.gradle
+   
+   repositories {
+     mavenLocal()
```
nmalfroy: If testing locally like this is something common, would it make sense to have this controlled via a gradle property?  Something like
```suggestion
     if ("${properties["localTest"]}" == "true") {
        mavenLocal()
     }
     ...
```

And you could then use `-PlocalTest=true` to run that way.

Anyways, just a thought

stairway/src/main/java/bio/terra/stairway/impl/StairwayThreadPool.java
okotsopoulos: Yes, that's the benefit that tipped the scale towards my usage of it over `ThreadPoolExecutor` (I liked that I was able to get rid of that custom tracking in this implementation as a result).  Tough to thread the needle sometimes between the diversity in tech stacks already present in our apps vs. what has become the prevailing standard (Spring Boot Java).

`ThreadPoolTaskExecutor` doesn't actually extend `ThreadPoolExecutor` either, so it would have been more involved (but not impossible) to support the non-Spring type.

stairway/src/main/java/bio/terra/stairway/impl/StairwayThreadPool.java
```diff
@@ -67,16 +48,4 @@ private void initializeFlightMdc(
     MdcUtils.addFlightContextToMdc(flightContext);
     MdcUtils.removeStepContextFromMdc(flightContext);
   }
-
-  protected void beforeExecute(Thread t, Runnable r) {
```
rtitle, line 71: I see these metrics were removed -- are these things we get "for free" with the switch to the Spring `ThreadPoolTaskExecutor`?

stairway/src/main/java/bio/terra/stairway/DefaultThreadPoolTaskExecutor.java
```diff
@@ -0,0 +1,31 @@
...
+    int poolSize = getPoolSize(maxParallelFlights);
+    super.setCorePoolSize(poolSize);
+    super.setMaxPoolSize(poolSize);
+    super.setKeepAliveSeconds(0);
+    super.setThreadNamePrefix("stairway-thread-");
+    super.initialize();
+  }
+
+  private int getPoolSize(Integer maxParallelFlights) {
+    if (maxParallelFlights == null || maxParallelFlights <= 0) {
```
nmalfroy, line 26: Just because it looks like you're switching using the syntax in StairwayImpl, this should probably be:
```suggestion
    if (Optional.ofNullable(maxParallelFlights).orElse(0) <= 0) {
```

But it's a whole character longer :)

rtitle (approved):
This looks good to me -- I like the approach of decoupling `StairwayThreadPool` from the executor (instead of inheriting from it).

stairway/src/main/java/bio/terra/stairway/impl/StairwayThreadPool.java
okotsopoulos: Happily, yes!  And they're more valuable to us as a metric rather than a debug log.  They're available as the metric `executor_active_threads`.

```
(base) okotsopo@wm111-e35 jade-data-repo % curl http://localhost:8080/actuator/prometheus | grep 'stairwayExecutor' | grep 'executor_active_threads'
…
executor_active_threads{name="stairwayExecutor",} 1.0
```

stairway/src/main/java/bio/terra/stairway/impl/StairwayImpl.java
```diff
@@ -105,6 +102,10 @@ public StairwayImpl(StairwayBuilder builder) throws StairwayExecutionException {
             ? "stairway" + UUID.randomUUID().toString()
             : builder.getStairwayName();
 
+    this.executor =
+        Optional.ofNullable(builder.getExecutor())
+            .orElse(new DefaultThreadPoolTaskExecutor(builder.getMaxParallelFlights()));
```
nmalfroy: This is maybe minor but I think it's slightly better practice to do
```suggestion
            .orElseGet(() -> new DefaultThreadPoolTaskExecutor(builder.getMaxParallelFlights()));
```

By using a supplier, you don't actually create the `DefaultThreadPoolTaskExecutor` until you need it.  It's a constructor so maybe not a huge deal but I wanted to call it out since I've been bit by this before.

stairway/src/main/java/bio/terra/stairway/impl/StairwayThreadPool.java
nmalfroy: Yeah, I had the same thought (and came to the same conclusion) for whatever that's worth :)

stairway/src/main/java/bio/terra/stairway/StairwayBuilder.java
```diff
@@ -30,7 +32,7 @@ public class StairwayBuilder {
 
   /**
    * Determines the size of the thread pool used for running Stairway flights. Default is
-   * DEFAULT_MAX_PARALLEL_FLIGHTS (20 at this moment)
+   * established in {@link DefaultThreadPoolTaskExecutor} (20 at this moment).
```
nmalfroy: Could just ponit to 
```suggestion
   * established in {@link DefaultThreadPoolTaskExecutor.DEFAULT_MAX_PARALLEL_FLIGHTS}
```

And I don't know that I'd put the value since that could get out of sync

stairway/src/main/java/bio/terra/stairway/impl/StairwayThreadPool.java
nmalfroy: Another note: would it make sense to have a backwards compatibly constructor that takes an int and constructs the `ThreadPoolTaskExecutor`. That would make this PR not a breaking change?  I could see the value in explicitly making it a breaking change to force people to configure a threadpool but wanted to check

nmalfroy (changes requested):
This looks great!  The reason I put in request changes is about some questions.  All minor but probably worthy of calling out

DEVELOPMENT.md
okotsopoulos: This was admittedly a copy-paste from the docs in terra-common-lib, I'll update to reflect the way TDR handles this (which is closer to what you've suggested).

stairway/src/main/java/bio/terra/stairway/impl/StairwayImpl.java
okotsopoulos: I'm glad you raised this, I hadn't made that connection and agree that it should be changed.  This is getting into territory where the optional chaining is hindering readability for me, so I'll revisit in full.

stairway/src/main/java/bio/terra/stairway/StairwayBuilder.java
okotsopoulos: I had tried the link like this and it didn't work, but I figured out how to get it to work… much better :)

Screenshot 2024-07-01 at 3 10 07 PM
Screenshot 2024-07-01 at 3 10 15 PM

stairway/src/main/java/bio/terra/stairway/impl/StairwayThreadPool.java
okotsopoulos: That's a good question, but `StairwayThreadPool`s are not public, and we don't support their direct implementation by consumers.

I don't see this as a breaking change as far as our contracts with consumers are concerned, but rather an implementation detail, so I'm not inclined to maintain that old constructor.

stairway/src/main/java/bio/terra/stairway/StairwayBuilder.java
nmalfroy: Ah good find!

stairway/src/main/java/bio/terra/stairway/impl/StairwayThreadPool.java
nmalfroy: OMG you're totally right.  Sorry, been looking at python too long and missed the absence of the `public` qualifier.

stairway/src/main/java/bio/terra/stairway/DefaultThreadPoolTaskExecutor.java
okotsopoulos: I'm actually going to walk back that syntax change, so I'll leave this line as it stands, thanks!

nmalfroy (approved):
You've answered all the questions I had :). Looks great!

stairway/src/main/java/bio/terra/stairway/DefaultThreadPoolTaskExecutor.java
nmalfroy: Absolutely.  Makes sense!

stairway/src/main/java/bio/terra/stairway/impl/StairwayThreadPool.java
okotsopoulos: No worries at all, I'm glad for the sanity check!

PR #152, opened 2024-07-11
[DCJ-507] Stairway with work queue enabled should be context-aware

okotsopoulos:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/DCJ-507

## Addresses

Context-aware Stairway previously assumed that the calling thread would have mapped diagnostic context (MDC) to persist to the child threads spawned for flight execution.

This was not the case when Stairway is configured with a work queue enabled (e.g. GCP Pub-Sub).  Here, flight information was written to the pub-sub topic without the calling thread's context, so when a message was processed and deserialized any logs emitted from that flight would be missing context set by the original calling thread (e.g. request ID).

Where we've felt this impact in TDR:
- Child flights submitted within the execution of a parent flight are missing the calling thread's context in its logs.
- While this is an uncommon pattern for flight execution, it is used when facilitating combined data ingests to datasets: the parent flight ingests tabular data, and child flights are spawned for each file being ingested.
- Much of our attention has been on debugging combined data ingests to Azure datasets: the inability to filter for all connected parent and child flight logs in GCP log console by request ID (`jsonPayload.requestId`) is slowing down investigative progress. 

## Summary of changes

- Added new helper method `MdcUtils.callWithContext` to call a callable and return its result with MDC's context map temporarily overwritten.
- Expanded `QueueMessageReady` to store calling thread context on construction.
- `QueueMessageReady.process()` sets the MDC using its stored calling thread context, which then makes the flight context-aware.

## Testing Strategy

- Added unit tests for `MdcUtils.callWithContext`.
- Expanded unit tests for `QueueMessageReady` with special attention to MDC behavior and propagation in serialization, deserialization, and processing.

~~Note: Sonarcloud coverage stats are incorrectly reporting 0% coverage.  They look to have never worked.  I do believe I've covered all of my changes in unit tests.~~ ETA: thanks @pshapiro4broad for fixing the Sonar scans, I pulled the fix into this branch and we are now seeing coverage stats.

### DIFF
--- a/stairway/src/main/java/bio/terra/stairway/queue/QueueMessageReady.java
+++ b/stairway/src/main/java/bio/terra/stairway/queue/QueueMessageReady.java
@@ -32,27 +32,26 @@ class QueueMessageReady extends QueueMessage {
 
   @Override
   public boolean process(StairwayImpl stairwayImpl) throws InterruptedException {
-    boolean processed = false;
-    // Save the initial thread context so that it can be restored
-    Map<String, String> initialContext = MDC.getCopyOfContextMap();
-    try {
-      MdcUtils.overwriteContext(callingThreadContext);
-      // Resumed is false if the flight is not found in the Ready state. We still call that
-      // a complete processing of the message and return true. We assume that some this is a
-      // duplicate message or that some other Stairway found the ready flight on recovery.
-      boolean resumed = stairwayImpl.resume(flightId);
-      logger.info(
-          "Stairway "
-              + stairwayImpl.getStairwayName()
-              + (resumed ? " resumed flight: " : " did not find flight to resume: ")
-              + flightId);
-      processed = true;
-    } catch (DatabaseOperationException ex) {
-      logger.error("Unexpected stairway error, leaving %s on the queue".formatted(flightId), ex);
-    } finally {
-      MdcUtils.overwriteContext(initialContext);
-    }
-    return processed;
+    return MdcUtils.callWithContext(
+        callingThreadContext,
+        () -> {
+          try {
+            // Resumed is false if the flight is not found in the Ready state. We still call that
+            // a complete processing of the message and return true. We assume that some this is a
+            // duplicate message or that some other Stairway found the ready flight on recovery.
+            boolean resumed = stairwayImpl.resume(flightId);
+            logger.info(
+                "Stairway "
+                    + stairwayImpl.getStairwayName()
+                    + (resumed ? " resumed flight: " : " did not find flight to resume: ")
+                    + flightId);
+            return true;
+          } catch (DatabaseOperationException ex) {
+            logger.error(
+                "Unexpected stairway error, leaving %s on the queue".formatted(flightId), ex);
+            return false;
+          }
+        });
   }
 
   public QueueMessageType getType() {
@@ -73,9 +72,5 @@ class QueueMessageReady extends QueueMessage {
 
   public Map<String, String> getCallingThreadContext() {
     return callingThreadContext;
-  }
-
-  public void setCallingThreadContext(Map<String, String> callingThreadContext) {
-    this.callingThreadContext = callingThreadContext;
   }
 }
--- a/stairway/src/main/java/bio/terra/stairway/queue/QueueMessageReady.java
+++ b/stairway/src/main/java/bio/terra/stairway/queue/QueueMessageReady.java
@@ -1,9 +1,12 @@
 package bio.terra.stairway.queue;
 
 import bio.terra.stairway.exception.DatabaseOperationException;
+import bio.terra.stairway.impl.MdcUtils;
 import bio.terra.stairway.impl.StairwayImpl;
+import java.util.Map;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
 
 /**
  * The Ready message communicates that a flight, identified by flightId, is ready for execution. By
@@ -16,6 +19,7 @@ class QueueMessageReady extends QueueMessage {
 
   private QueueMessageType type;
   private String flightId;
+  private Map<String, String> callingThreadContext;
 
   private QueueMessageReady() {}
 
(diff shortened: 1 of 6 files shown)


rushtong (approved):
This looks reasonable to me 👍🏽

stairway/src/main/java/bio/terra/stairway/queue/QueueMessageReady.java
```diff
@@ -59,4 +70,12 @@ public String getFlightId() {
   public void setFlightId(String flightId) {
     this.flightId = flightId;
   }
+
+  public Map<String, String> getCallingThreadContext() {
```
pshapiro4broad, line 73: It would be nice to not have to add public APIs that are only used in tests.

For `getCallingThreadContext()`, could the test instead assert on `MDC.getCopyOfContextMap()`?

For `setCallingThreadContext()`, could the test instead call `MDC.setContextMap()` before constructing `QueueMessageReady`?

stairway/src/main/java/bio/terra/stairway/queue/QueueMessageReady.java
```diff
@@ -37,11 +46,13 @@ public boolean process(StairwayImpl stairwayImpl) throws InterruptedException {
               + stairwayImpl.getStairwayName()
               + (resumed ? " resumed flight: " : " did not find flight to resume: ")
               + flightId);
-      return true;
```
pshapiro4broad, line 40: Was this change necessary? If it works either way, I think returning a boolean directly is clearer than creating state which must be returned at the end.

stairway/src/main/java/bio/terra/stairway/queue/QueueMessageReady.java
```diff
@@ -16,18 +19,24 @@ class QueueMessageReady extends QueueMessage {
...
+    this.callingThreadContext = MDC.getCopyOfContextMap();
   }
 
   @Override
   public boolean process(StairwayImpl stairwayImpl) throws InterruptedException {
+    boolean processed = false;
+    // Save the initial thread context so that it can be restored
+    Map<String, String> initialContext = MDC.getCopyOfContextMap();
     try {
+      MdcUtils.overwriteContext(callingThreadContext);
```
pshapiro4broad: Could this be done inside `McdUtils`? Something like
```
    MdcUtils.callWithContext(callingThreadContext, () -> {
        ...
        stairwayImpl.resume(flightId)
        ...
    });
```

I think it would be better to expose less of the inner workings of `MdcUtils` if possible. And there may be other cases where we want to overwrite the context temporarily like this.

stairway/src/test/java/bio/terra/stairway/queue/QueueMessageTest.java
```diff
@@ -1,32 +1,106 @@
...
+class QueueMessageTest {
+
+  @Mock private StairwayImpl stairway;
+  private static final String FLIGHT_ID = "flight-abc";
+  private static final Map<String, String> CALLING_THREAD_CONTEXT =
+      Map.of("requestId", "request-abc");
+
+  @BeforeEach
+  void beforeEach() {
+    MDC.clear();
```
pshapiro4broad, line 37: Would it be better to use a static mock instead of the actual MDC API? We don't use mocks yet for MDC, so it may be too big a change for this PR, but it's something to consider if we'd like to isolate our tests from the MDC implementation, and might make it easier to check if our code is doing what we expect.

pshapiro4broad (approved):
Looks OK, although I'd prefer making fewer APIs public for tests if possible

stairway/src/main/java/bio/terra/stairway/queue/QueueMessageReady.java
okotsopoulos: Good points.

> For getCallingThreadContext(), could the test instead assert on MDC.getCopyOfContextMap()?

No, because the MDC written to the `QueueMessageReady` object on its construction can differ from the current value of the MDC.  I can remove the `public` modifier and annotate as visible for testing, though.

> For setCallingThreadContext(), could the test instead call MDC.setContextMap() before constructing QueueMessageReady?

Yes it could, and then I think I could remove this method.

stairway/src/main/java/bio/terra/stairway/queue/QueueMessageReady.java
okotsopoulos: Because I added a `finally` clause to restore the thread's context to its original state, I do think this was a necessary change.

stairway/src/test/java/bio/terra/stairway/queue/QueueMessageTest.java
okotsopoulos: Hmm, yeah this feels like too big a change for this PR.  But you make good points about the benefits of test isolation.

stairway/src/main/java/bio/terra/stairway/queue/QueueMessageReady.java
okotsopoulos: Ah, but it's possible that this won't be necessary if I make a change like you described here -- https://github.com/DataBiosphere/stairway/pull/152/files/93f139fd72b2e1d0b82000676a5a163b8130236e#r1676240309

I'll take a closer look.

pshapiro4broad: > Because I added a finally clause to restore the thread's context to its original state, I do think this was a necessary change.

Even if you `return` inside a `try`, the `finally` will still run.

stairway/src/main/java/bio/terra/stairway/queue/QueueMessageReady.java
okotsopoulos: Thanks for the suggestion. I added new public utility method `MdcUtils.callWithContext` and removed the `public` modifier on `MdcUtils.overwriteContext`.  Any possible refactoring of other code to use this method I'll call out of scope.

stairway/src/main/java/bio/terra/stairway/queue/QueueMessageReady.java
okotsopoulos: I reverted to the old return strategy.

stairway/src/main/java/bio/terra/stairway/queue/QueueMessageReady.java
okotsopoulos: Update: I was able to remove the setter in full, but I did need to leave the getter as public so that serialization and deserialization would work.  This was caught by a unit test failure when I removed the public modifier, so that's nice :)

Other preexisting setters can likely be removed as well, but I'll designate that work as OOS.

fboulnois (approved):
looks good 👍

PR #126, opened 2024-02-29
[DSEC-554] sonar scan

TomConner:
Adding sonar scan to build and test workflow.

PR #153, opened 2024-07-12
PF-3003: Add code coverage to sonar report

pshapiro4broad:
Generate and push code coverage results to the repo's sonar scan.

Sonar support was added in https://github.com/DataBiosphere/stairway/pull/126, but since the build wasn't generating coverage data there wasn't any coverage in the sonar scan.

I generated a sonar coverage report by running locally, at PR creation time the results are here: https://sonarcloud.io/component_measures?id=DataBiosphere_stairway&metric=coverage&view=list

Note that overall coverage is only 66%, but that number is skewed by the `stairctl` subproject which has no unit tests.

Also note, the coverage report for this PR will be empty since sonar only shows coverage for files included in the current commit, and no java code files were changed in this PR.

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -11,9 +11,9 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Set up JDK
-        uses: actions/setup-java@v3
+        uses: actions/setup-java@v4
         with:
           java-version: '17'
           distribution: 'temurin'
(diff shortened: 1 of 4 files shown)


mmorgantaylor (approved):

.github/workflows/build-and-test.yaml
```diff
@@ -11,9 +11,9 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
```
pshapiro4broad, line 14: drive-by change, updated these to the current versions

okotsopoulos (approved):
Thank you!

> Note that overall coverage is only 66%, but that number is skewed by the stairctl subproject which has no unit tests.

Very much an aside / out of scope but I am personally in favor of removing stairctl subproject: it doesn't work and I don't think we have any intention of investing in it.

pshapiro4broad:
> Very much an aside / out of scope but I am personally in favor of removing stairctl subproject: it doesn't work and I don't think we have any intention of investing in it.

I created a ticket for the work to remove stairctl (or fix it, if we have a case where it might be useful):  https://broadworkbench.atlassian.net/browse/PF-3004

PR #174, opened 2025-08-04
skipJreProvisioning for Sonar

davidangb:
Configure the Sonar scanner to use the pre-existing Java installation instead of provisioning its own.

This change appears to resolve an execution problem in #170. See #173, which is a copy of #170 but with this flag enabled.

I hope to get this PR approved and merged, then ask Dependabot to rebase #170, then merge #170.


rushtong (approved):

fboulnois (approved):
👍

